PR #60, opened 2018-03-11
Fix for news tests

zootos:


zootos:
Closed due to problems with selecting the wrong base and changing it

PR #77, opened 2018-04-05
Make img and aimg tags links when they have a href attribute

kalkins:


zootos:
I do not see the point of the `aimg[href]` selector. `aimg` is not a valid HTML tag, and is not used in our project. 

Also, does the script have to be included in the base header?  While it is a very small file, it still has to be loaded for every page. It is only really used in a single page and could rather be included in the `head` block in that template.

kalkins:
Amund added the aimg tag on the front page and article list page, with the intention of replacing the a tag, as it caused some visual problems. He didn't write the logic for it, so the tag has been laying dormant for some time.

As I've made it work for img tags with the href attribute the aimg tags can probably be replaced, but it might involve changing some css.

Right now the script is used on the front page, and the list pages for events and articles. I loaded it in the base header because it's so small it doesn't affect performance, and it could conceivably be used on other pages in the future.

kalkins:
The aimg and atxt tags have been replaced with img and div, and the aimg.js file was renamed.

The `nth-of-type` in css selectors was replaced with django tags

PR #94, opened 2018-04-18
Add 404 page

Henfau:


kalkins:
There is a couple of problems with this:
* Don't add local_settings.py to git. Also, the file should be "local_settings.py" not ".local_settings.py"
* It looks like you added a view, then removed it but committed the leftover changes in the file
* There is a test that fails
* The pull request should be to dev, not master

I'll close this request. You can create a new when you have fixed this

PR #99, opened 2018-04-20
404

Henfau:


kalkins:
Ehh... As you managed to delete the settings file I would suggest rebasing the commits. Also, I would prefer git messages to be more descriptive and not include profanity

PR #100, opened 2018-04-20
404

Henfau:


kalkins:
Looks good. Good commit messages, though "Add empty line between ... and ..." might be too verbose. Newlines aren't that important to document

PR #133, opened 2018-10-09
Committee edit

kalkins:


kalkins:
There isn't a convenient way to get to the committee list yet, but I wasn't sure how that should be done

PR #137, opened 2018-10-14
Fix for navbar user dropdown on mobile.

zootos:
Fixes two issues with the user dropdown on mobile.
- The dropdown is now initially closed
- `.dropdown()` was called on the dropdown icon, creating an extra div

Also fixes an issue, where Cmd/Ctrl clicking a link in the dropdown would change the dropdown text (the name), to the text of the selected link.

### DIFF
--- a/web/templates/web/header.html
+++ b/web/templates/web/header.html
@@ -35,8 +35,7 @@
         </a>
         {% if user.is_authenticated %}
             <div class="ui dropdown item">
-                <div class="text">{{ user.first_name|first_word }}</div>
-                <i class="dropdown icon"></i>
+                {{ user.first_name|first_word }} <i class="dropdown icon"></i>
                 <div class="menu">
                     {% if user|has_any_permissions or user.is_superuser %}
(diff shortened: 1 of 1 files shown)


ddabble:
Getting a `Dropdown: The method you called is not defined. ["nothing"]` error, from `$('#nav-user-dropdown').dropdown(action="nothing")` in `header.html:93`..?

zootos:
@ddabble I have fixed that now. I did some testing with other actions than `select` and must have forgotten to switch back when staging.

ddabble:
Hm, the items in the dropdown still get marked as selected, which looks a bit weird.. Also, think I found the proper way to pass the action argument, pushing now:

ddabble:
Is this okay? :)

zootos:
Yeah, I think so. The dropdown remains open, which seems weird at first. However, it might actually be better to leave it open than close it.

ddabble:
Yeah, agree with that latter point, as it seems to be the norm on other websites 👍 Merging!

PR #147, opened 2018-11-03
Reservation system upgrades

zootos:
These upgrades rewrite much of the model code, and thus require some manual data migration.

Contains several major upgrades to the reservation system (Also fixes several small issues):
- Rules pages for each machine type
- Rules for maximum reservation time for variable time periods
- Upgraded quotas that allow a given user to create more reservations at a time or ignore rules for a given number of reservations
- The end time of a reservation can now be changed after the reservation has started, including a "Done" button to set the end time to the current time.
- Registration of 3D-printer course participation on the website, including downloadable xlsx lists.


kalkins:
Never mind, I had a template crash

zootos:
What template crashed and in what case?

kalkins:
That fixed it. I think it's ready to merge

PR #165, opened 2018-11-16
Added a message in quota panel when user has no quotas

hermabe:


make_queue/templates/make_queue/quota/quota_user.html
```diff
@@ -23,4 +23,11 @@
...
         </div>
     </div>
-{% endfor %}
\ No newline at end of file
+{% endfor %}
+
+{% if user_quotas|length == 0 %}
+	<div class="noquota">
+		{% trans 'No quotas' %}
+	</div><br>
```
zootos: Do not see the point of using the `<br>` tag here, as it does nothing.

make_queue/templates/make_queue/quota/quota_user.html
```diff
@@ -23,4 +23,11 @@
             {% endif %}
         </div>
     </div>
-{% endfor %}
\ No newline at end of file
+{% endfor %}
+
+{% if user_quotas|length == 0 %}
```
zootos: Use djangos [for-empty](https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#for-empty) instead.

make_queue/static/make_queue/css/quota_panel.css
```diff
@@ -0,0 +1,5 @@
+.noquota {
```
zootos: No need to add custom CSS for this. Can use the `<i>...</i>` tag for italic and there is a global CSS class named `text-centered` for centering text.

make_queue/templates/make_queue/quota/quota_user.html
```diff
@@ -23,4 +23,11 @@
...
+	<div class="noquota">
+		{% trans 'No quotas' %}
```
zootos: Rename to "User has no personal quotas"/"Brukeren har ingen personlige kvoter"?

PR #167, opened 2018-11-20
Design/next reservation time

hermabe:
When a machine is available, the time until next reservation will be shown.


zootos:
This pull request should probably be into `dev`, not `design/reservation`

zootos:
Waiting for #161 to be reviewed before reviewing this, as this is directly dependent on #161

PR #168, opened 2018-11-20
Freshen up navbar

ddabble:
* Moved language link and user dropdown to the right edge of the window
 * Moved "About us" to navbar and removed "Home"
 * Increased font size
 * Added bubble backgrounds with MAKE's primary colors to each element on the navbar
 * Centered navbar in desktop layout, and centered logo in mobile layout
 * Moved burger menu button to the right of the logo, instead of left


web/templates/web/header.html
```diff
@@ -6,78 +6,92 @@
...
-    <div class="ui secondary inverted menu">
-        <a class="item" href="{% url 'front-page' %}">
-            {% trans "Home" %}
+
+    <div id="burger">
+        <i class="sidebar icon"></i>
+    </div>
+
+    <nav id="nav" class="ui secondary inverted menu">
+        <a class="nav-item item" href="{% url 'events' %}">
```
zootos: I do not really see the need for the class `nav-item`. I personally find using the selector `#nav .item` to be just as descriptive and much easier to read than `#nav .nav-item`.

web/templates/web/header.html
```diff
@@ -6,78 +6,92 @@
...
                     {% if user|has_any_permissions or user.is_superuser %}
-                        <a class="item" href="{% url 'adminpanel' %}">
-                            {% trans "Administration" %}
+                        <a class="dropdown-item item" href="{% url 'adminpanel' %}">
+                            <div class="dropdown-item-text">{% trans "Administration" %}</div>
                         </a>
                     {% endif %}
-                    <a class="item" href="{% url 'profile' %}">
-                        {% trans "Profile" %}
+                    <a class="dropdown-item item" href="{% url 'profile' %}">
```
zootos: Same as above for `nav-item`

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
 
-.lang-form {
-    display: none;
+#header #nav .nav-item .bubble-background,
+#header #side-nav .nav-item .bubble-background {
+    position: absolute;
+    width: 100%;
+    height: var(--font-size);
+    border-radius: 2em;
+    z-index: 1;
```
zootos: Is the z-index here needed? I would think that the only needed z-index is the one for the text, so that it overlaps the bubble.

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
-@media screen and (max-width: 900px) {
-    .limg.large {
+#header #side-nav .nav-item {
+    margin-right: 0;
+}
+
+#header #side-nav #nav-user-dropdown .menu {
+    background-color: rgb(34, 43, 52);
+    border-radius: 7px;
+    /* For some reason doesn't work with `border: 2px rgb(220, 221, 222)` */
```
zootos: You have to specify `border: 2px solid rgb(220, 221, 222)`

web/templates/web/header.html
```diff
@@ -6,78 +6,92 @@
...
-    <div class="burger">
-        <i class="sidebar icon"></i>
-    </div>
+<header id="header">
+    <div class="logo-centering">{# Empty div for centering logo (on mobile), specified with grid-template-columns #}</div>
+
     <div class="logo">
+        <div>{# Empty div for spacing left of logo, specified with grid-template-columns #}</div>
         <a href="{% url 'front-page' %}">
             <img class="limg large" src="{% static "web/img/logo_white.svg" %}">
```
zootos: Do not see the need for the `limg` class here, it is only used for a selector, which actually does not need it.

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
-        display: inline;
+    #header .logo .limg.small {
+        display: none;
     }
 }
 
+/* Switch to small logo */
 @media screen and (max-width: 340px) {
-    .limg.medium {
+    #header .logo .limg.large {
```
zootos: For all of these logo selectors, it would probably be easier to specify that the `display` property should be set to `block` in intervals, rather than having to use specify `none` on all of the others.

web/templates/web/header.html
```diff
@@ -6,78 +6,92 @@
...
+        </div>
+
     </div>
-</div>
+</header>
 
-{# This form is hidden, but is submitted through JQuery #}
-<form class="lang-form" action="{% url 'set_language' %}" method="POST">
+{# This form is hidden, but is submitted through jQuery #}
+<form id="lang-form" action="{% url 'set_language' %}" method="POST" style="display:none">
```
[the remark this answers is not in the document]
ddabble: Just moved `display: none` from the stylesheet, because I felt it didn't really belong there, and to make it more explicit. Should I remove the [submit button](https://github.com/MAKENTNU/web/blob/wip/navbar/web/templates/web/header.html#L103) instead?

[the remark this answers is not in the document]
zootos: You can just remove the submit button and the `style="display: none"`. The submit button is not used (it is hidden) and since all of the input fields are hidden, nothing will be displayed.

web/static/web/css/style.css
```diff
@@ -4,8 +4,8 @@ html, body {
 
 .wrapper {
     display: table;
-    height: 100%;
     width: 100%;
+    height: 100%;
```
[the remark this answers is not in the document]
ddabble: Just because width comes before height elsewhere in the stylesheet, because, you know, (x, y) :p

web/static/web/css/header.css
ddabble: Oh right, thanks! :)

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
 }
 
-.limg.large {
-    display: inline;
+#header .logo .limg.small {
+    display: none;
 }
 
-.burger {
+#header #burger {
```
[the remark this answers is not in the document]
ddabble: That's true. I did it purely because I thought it increased readability, like the (unnecessarily) long chained selectors elsewhere in the stylesheet, because you can get an overview of which elements are inside which just by reading the selectors, and because I think it's easier to find where the properties for a specific element is located as I scroll through the stylesheet.

For example, if I was trying to find `.item` inside `#nav` inside `#header`, it would be quicker to be able to scroll down until I found `#header`, then `#header #nav`, and finally a little further to find `#header #nav .item`. I guess the alternative would be to sort the selectors alphabetically, or to only include the necessary selectors and place them in the same order as they appear in the HTML, and relying on either remembering that exact order to find a specific selector, or using Ctrl+F - which I guess is an absolutely feasible way of doing it :p

I do, however, get kind of blind to the readability after having read the same code over and over, so I'm unsure whether it's as readable as I initially thought, now that you pointed it out.

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
 
-.burger {
-    align-self: center;
+#header #burger i {
+    color: #FFFFFF;
+    font-size: 30px;
+}
+
+#header #nav {
     justify-self: center;
```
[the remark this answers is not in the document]
ddabble: It makes the navbar not take up the entire width of its parent (with mobile layout), which most noticably makes the buttons inside not cover the entire page. Can move to `@media screen and (max-width: 991px)` for clarity. Or should I replace it with something more explicit?

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
 
-.burger:hover {
+#header #burger:hover {
     cursor: pointer;
 }
 
-.burger {
-    align-self: center;
+#header #burger i {
+    color: #FFFFFF;
```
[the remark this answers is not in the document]
ddabble: Thanks!

web/static/web/css/header.css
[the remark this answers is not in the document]
zootos: Moving it to `@media screen and (max-width: 991px)` sounds good

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
+    font-size: 30px;
```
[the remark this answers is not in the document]
ddabble: Ooh, great, thanks :D

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
 
-.burger i {
-    color: #FFFFFF;
-    font-size: 30px;
+#header #nav .nav-item,
+#header #side-nav .nav-item {
+    height: calc(var(--font-size) + 2em);
+    margin: calc(var(--nav-space-between) / 2);
+    padding: 0;
+    border-radius: 0;
```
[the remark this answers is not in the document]
ddabble: True, don't really know why it's there.. Will remove.

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
+#header #nav .nav-item:nth-child(1):hover .bubble-background {
+    background-color: rgba(239, 184, 14, 0.9);
+}
+
+#header #nav .nav-item:nth-child(2):hover .bubble-background {
+    background-color: rgba(0, 235, 179, 0.8);
+}
+
+#header #nav .nav-item:nth-child(3):hover .bubble-background {
+    background-color: rgba(12, 100, 141, 1); /* MAKE's blue primary color (12, 66, 106), with brightness turned up from 149 to 114 */
```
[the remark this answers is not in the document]
ddabble: Because it was too dark on its own, relative to the background and the other colors. Like the comment says, it's the same color as in the graphic profile manual, just with brightness turned up (which technically doesn't make it the same color, I know :p ), and I think that it still looks like it fits, which is the most important thing, I think.

web/static/web/css/header.css
ddabble: Oh, yes, that's true! It was a remnant from earlier in the development, when the bubble-backgrounds used to overlap each other.

web/static/web/css/header.css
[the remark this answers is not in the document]
zootos: I get what you are trying to achieve. However, I see a couple of problems.

- Duplication of code. The same X number of elements follow each other all the time. Meaning that you have these long chains of selectors all over the place (much code). The code is also kind of duplicated across HTML and CSS.
- A simple HTML change may require changes to most of the CSS, if it changes some of the ordering of the elements. Rather than only requiring changes to a couple of selectors.
- Long selectors makes the code look more complex. There is a reason why one usually set a limit to the length of line of code.

Personally, I find it just as easy to find the correct location in the CSS file if you instead group selectors based on their elements. E.g. everything related to `#nav` is grouped. Then everything related to `.item` is grouped and so on. That is, group selectors based on which class/id they change. 

I think sorting selectors alphabetically would be much harder to use. Especially due to the nature of CSS selectors. In which selectors for a single element are not necessarily grouped together alphabetically.

web/static/web/css/header.css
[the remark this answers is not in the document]
zootos: It is probably fine to use here.

web/static/web/css/header.css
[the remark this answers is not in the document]
ddabble: Grouping selectors is basically what I meant when I said "place them in the same order as they appear in the HTML" :p I agree with that last point on sorting alphabetically. Other than that, you bring up some good points! Will change accordingly.

web/templates/web/header.html
ddabble: Mostly because there are `.item`s within `#nav` that should not be selected, and because the `.nav-item`s are very specific "items", in contrast to the ubiquitous `.item`. Can fix it with `#nav > .item`, though, or a change of name, like `.nav-button`. What do you think?

zootos: I think `#nav > .item` is good. I tells me that all items in the upper level of the navigation bar should be selectable (etc.).

web/static/web/css/header.css
ddabble: Not really sure what you mean by "set to `block` in intervals".. Specifying `none` for the other image sizes is purely for readability, because then you get the same format wherever the size is changed.

web/static/web/css/header.css
```diff
@@ -1,144 +1,332 @@
...
+}
+
+#header #nav .nav-item .nav-item-text,
+#header #side-nav .nav-item .nav-item-text {
+    z-index: 3;
+    transition: color var(--dur) ease;
+}
+
+#header #nav .nav-item:nth-child(1):hover .bubble-background {
+    background-color: rgba(239, 184, 14, 0.9);
```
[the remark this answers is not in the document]
ddabble: Nice suggestion! Don't know why I didn't think of that :p Will change to using classes for the colors instead. Unsure what you mean by "set the color just on hover", though..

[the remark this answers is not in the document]
zootos: Now, you just set the color on hover, since `.bubble-background` has a width of 100% when not hovered (for the animation). Meaning that if you just make a class, e.g:
```css
.make_bg_yellow_80 {
    background-color: rgba(239, 184, 14, 0.8);
}
```

and doing `... class="make_bg_yellow_80 bubble-background"...` would not work correctly. And it is nice to have generic classes for the colors in the profile manual (graphical manual)(Eventually all of the colors in it should be included in the index.css file, with respective variables as well). 

If you make `.bubble-background` have `visibility: hidden` when not hovered, and adding the transition `visibility var(--dur) ease-out` you can have the background color set even when not hovered.

web/static/web/css/header.css
ddabble: Very good idea! Made it much more readable :)

web/static/web/css/header.css
zootos: Might want to remove the comment as well

web/static/web/css/header.css
```diff
@@ -1,144 +1,274 @@
...
-        display: none;
+    #header div.logo-centering {
+        display: inline;
     }
 
-    .nav {
-        grid-template-columns: 1fr 3fr;
+    #header .logo {
+        grid-template-columns: 1em var(--logo-space) 1em;
     }
```
zootos: You could just remove this. For screen sizes larger than 992px, `grid-template-columns` is set to `var(--logo-padding) var(--logo-space) minmax(var(--nav-space-between), var(--logo-padding))` which due to the redefinition of `--logo-padding` to `1em` for screen sizes smaller than or equal to `1100px` (line 165 - 170), is equivalent to the definition here.

web/static/web/css/header.css
```diff
@@ -1,144 +1,274 @@
...
-    }
+#nav > .item .make_bg_yellow.bubble-background {
+    filter: brightness(90%);
+}
 
-    .limg.small {
-        display: none;
-    }
+#nav > .item .make_bg_turquoise.bubble-background {
+    filter: brightness(90%);
```
zootos: Combine with the block above?

web/static/web/css/header.css
```diff
@@ -1,144 +1,274 @@
...
 
+    #nav > .item,
+    #side-nav > .item {
+        margin: 0;
     }
 
-    .nav.active {
-        height: auto;
+    #side-nav {
+        align-self: end;
```
zootos: Does not seem to have an effect here, as this is the only content in this row of the grid.

web/static/web/css/header.css
```diff
@@ -1,144 +1,274 @@
...
+#side-nav .lang-link .link .text {
+    margin-right: 0.5em;
+}
+
+#side-nav .lang-link .link .globe.icon {
+    margin: 0;
+}
+
+/* Narrower space between elements in navbar */
+@media screen and (max-width: 1100px) {
```
zootos: Could we use `1200px` instead, so that the change happens at the same time as the change of logo type?

web/static/web/js/dropdown.js
```diff
@@ -1,6 +1,6 @@
 $(function () {
-    $('.burger').click(function () {
-        $('.nav').toggleClass('active');
+    $('#burger').click(function () {
+        $('#header').toggleClass('active');
     });
```
zootos: Think that you could just delete this file instead of changing it. From what I could see, it is not used anywhere in the project.

web/static/web/css/header.css
ddabble: I'd prefer not, because it was just a coincidence that both yellow and turquoise worked with the same brightness level; they should have independent properties.

web/static/web/css/header.css
ddabble: True, I think I kept it there to signal that `#side-nav` in principle should stick to the bottom of the expanded burger menu, if the header for some reason got its height increased, or something.. I'd like to keep it there because of that, but might as well remove it :)

web/static/web/css/header.css
ddabble: I'd rather keep it - or optimally change it to `var(--logo-padding) var(--logo-space) var(--logo-padding)` for verbosity, to not make it dependent of `var(--nav-space-between)`; because if the latter variable were to increase, the logo would get offset from the center.

zootos: I see the point of not being dependent on `var(--nav-space-between)`. I feel that the explicit declaration you mention, `var(--logo-padding) var(--logo-space) var(--logo-padding)`, would be better for readability.

web/static/web/css/header.css
zootos: Probably just as good to keep it there.

web/static/web/css/header.css
ddabble: True, it looks a bit weird.. Tried removing `--logo-padding: 1em` at [line 170](https://github.com/MAKENTNU/web/pull/168/files#diff-ce9c358c5c2fdf54d88770e2901ab361R170), and it looks a little better. (The logo still gets larger, though - because it's given more space when `--nav-space-between` narrows.) Looks good with your initial suggestion, too. What do you think? Should the sizing be different?

zootos: Personally, I prefer the my solution. The navbar does get a bit more cramped. However, it feels more natural when reducing the size of my window (e.g if I were to place two windows side by side on my computer screen), due to it being a smoother transition (that is no abrupt change).

ddabble: So replace `1100px` with `1200px` and remove `--logo-padding: 1em`..?

zootos: Sounds good

ddabble: Nice, no worries :)

PR #173, opened 2018-11-22
Fix/admin reservations

hermabe:
Ordered MAKE NTNU reservations by start time in admin panel.
Also changed the title so it doesn't say "My reservations" in the admin panel.


make_queue/views/admin/reservation.py
```diff
@@ -13,4 +13,4 @@ def get_context_data(self):
 
         :return: The context required for the view
         """
-        return {"reservations": Reservation.objects.exclude(event=None, special=False), "admin": True}
+        return {"reservations": Reservation.objects.exclude(event=None, special=False).order_by("start_time"), "admin": True}
```
zootos: Would it make more sense to order them by latest start time first? As is done for the user reservations. Otherwise something has to be done about the ordering when showing previous reservations. These will now be shown first.

hermabe: Users are most likely interested in the reservations closest in time. For future and past reservations this will result in an opposite ordering. I think most people are interested in upcoming reservations and rarely look at past reservations.

I agree that the list will look a bit odd when showing all reservations, as the top ones could be from a long time ago and no longer relevant. Can change to reverse chronological order for consistency with the page for personal reservations.

zootos: I agree that users are interested in the reservations closest in time. However, I feel that keeping the consistency between the personal and admin reservations page would be good for now. 

In the long term a redesign of the page (both personal and admin) is probably a good idea. Personally, I feel that the page is quite messy (my own bad), and think that the same information could be presented in a clearer way. Additionally, the admin page could do with a better representation of the different types of reservations. It also think it would be nice if the admin page could provide an overview of who is currently using which machine. However, this is  not relevant to this PR.

PR #174, opened 2018-11-30
Added LDAP-search functionality

hermabe:
On the course registration page, the full name field autofills when username is entered.
Also added a function for finding user info from email.


make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -81,9 +81,18 @@ <h1>
...
         ;
-        $(".clear-dropdown").click(function() {
+        $(".clear-dropdown").click(function () {
             $(this).closest("div.ui.field").find("div.dropdown").dropdown("clear");
         });
+
+        $("input[name='username']").focusout(function () {
+                $.ajax(langPrefix + "/reservation/json/" + $("input[name='username']").val() + "/", {
+                    success: function (data) {
+                        $("input[name='name']").val(data["full_name"]);
```
zootos: Could we add a [warning message](https://semantic-ui.com/collections/message.html#warning) when the user cannot be found as well (In similar fashion to the one shown when adding a user)? With text like "The given username could not be found in NTNUs system. Are you sure you entered the correct username?". Would solve many of the problems we currently have with usernames sometimes being entered incorrectly.

make_queue/util/user_info.py
```diff
@@ -0,0 +1,53 @@
+from django.contrib.auth.models import User
+import ldap
+
+LDAP_HOST = 'ldap://at.ntnu.no'
```
zootos: Could we move this file to the `dataporten` app with a file name of `ldap.py`? It feels more natural to collect all the external user information sources in one app. Additionally, adding [docstrings](https://www.python.org/dev/peps/pep-0257/) would be great.

make_queue/util/user_info.py
```diff
@@ -0,0 +1,53 @@
...
+
+def get_user_details_from_email(email):
+    user = User.objects.filter(email=email).first()
+    if user:
+        return get_user_details_from_username(username=user.username)
+
+    ldap_results = LDAP_search("mail", email)
+    if len(ldap_results) == 0:
+        return None
+    return get_user_details_from_username(get_LDAP_field(ldap_results, "username"))
```
zootos: This could result in two searches to the database, while the information should be available after the first search. Would be better to move the retrieval of data from the LDAP results to a helper function and instead call that.

make_queue/util/user_info.py
```diff
@@ -0,0 +1,53 @@
...
+    return user_details
+
+
+def get_user_details_from_email(email):
+    user = User.objects.filter(email=email).first()
+    if user:
+        return get_user_details_from_username(username=user.username)
+
+    ldap_results = LDAP_search("mail", email)
+    if len(ldap_results) == 0:
```
zootos: Invert to
```python3
if ldap_results:
    return get_user_details_from_username(get_LDAP_field(ldap_results, "username"))
return {}
```
as that is more readable.

make_queue/util/user_info.py
```diff
@@ -0,0 +1,53 @@
...
+        if len(ldap_results) == 0:
+            return None
+        user_details["full_name"] = get_LDAP_field(ldap_results, "full_name")
+        user_details["email"] = get_LDAP_field(ldap_results, "mail")
+
+    return user_details
+
+
+def get_user_details_from_email(email):
+    user = User.objects.filter(email=email).first()
```
zootos: Would be great if one could specify to retrieve the data from the LDAP server even though the user exists locally. E.g. if I want to find the NTNU username and not the derived username we get through the email supplied from dataporten. Could probably just use a kwarg to indicate if we want to use cached (local data). 

```python3
def get_user_details_from_email(email, use_cached=True):
```

A similar feature would also be great for the username search.

make_queue/util/user_info.py
```diff
@@ -0,0 +1,53 @@
...
+def LDAP_search(field, value):
+    l = ldap.initialize(LDAP_HOST, bytes_mode=False)
+    l.simple_bind()
+    query = "({}={})".format(LDAP_FIELDS[field], value)
+    return l.search_s(LDAP_BASE, ldap.SCOPE_SUBTREE, query)
+
+
+def get_LDAP_field(ldapData, field):
+    return ldapData[0][1].get(LDAP_FIELDS[field], [b''])[0].decode()
+    # list<tuple<str, dict<str, list<bstr>>>>
```
zootos: I think it would be more useful with a description of the data structure (how is everything packed) we receive from LDAP, rather than just the datatypes. E.g. "We get a list of results. These consists of tuples of...".

make_queue/util/user_info.py
hermabe: Have added the option to only use LDAP search. Can also use the `get_user_details_from_LDAP(search_field, search_value)` function directly.

make_queue/util/user_info.py
hermabe: Moved to `dataporten/ldap_utils.py` to not shadow the `python-ldap` package.

make_queue/templates/make_queue/course/registration_edit.html
hermabe: Added a large error message at the top of the page when the user is not found. The users should also be able to verify that the username entered matches with the full name field, so hope this becomes lesser of a problem in the future.

make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -25,6 +25,15 @@ <h1>
         </h1>
 
         {% block messages %}
+            <div id="messages" style="margin-bottom:20px; display: none;">
+                <div class="ui red message">
+                    <i class="close icon"></i>
```
zootos: I think the outer div is not needed. If you remove it, the `message` class already has a margin on the bottom. And you can use `ui hidden red message` as the class, to hide the message initially. Then just us an id on the `ui red message` div to easily select it with jQuery.

Might also be better to use `warning` instead of `red`, as this is not necessarily an error (Dataporten is not limited to NTNU users only, however, LDAP is).

make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -73,17 +82,24 @@ <h1>
...
-            .on('click', function () {
-                $(this)
-                    .closest('.message')
-                    .transition('fade')
-                ;
-            })
-        ;
-        $(".clear-dropdown").click(function() {
+        $('#messages').click(function () {
+            $(this).fadeOut();
```
zootos: This change away from `$('.message .close')...` means that the message for successful adds will not be closable. So you could probably just revert this change and everything (with the adjustments from the other comment) should work just fine.

hermabe: I completely forot about the confirmation message. Have fixed it now.

dataporten/ldap_utils.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+    if use_cached:
+        user = User.objects.filter(email=email).first()
+        if user:
+            return {
+                "username": user.username,
+                "full_name": user.get_full_name(),
+                "email": email,
+            }
+    return get_user_details_from_LDAP("mail", email)
```
zootos: This errors. Should probably be `"email"`.

hermabe: Yes, forgot to change that.

PR #177, opened 2018-12-07
Set autofocus on username field on course registration page

hermabe:


make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -102,5 +102,9 @@ <h1>
                 },
             });
         });
+
+        $("input[name='username']").ready(function () {
+            $("input[name='username']").focus();
+        });
```
zootos: There is a more Django way of doing this. Instead of defining it in Javascript, you can define it for the form in the Django code. That is, you can add 
`"username": forms.TextInput(attrs={"autofocus": "autofocus"})` to the widgets for the Printer3DCourseForm in the `make_queue/forms.py`

PR #179, opened 2018-12-10
Multilingual input

zootos:
Adds support for multilingual input fields and fixes minor issues with content boxes.

- Add multilingual input fields with form fields and widgets for TextInput, Textarea, RichText (CKEditor) and RichTextUploading (CKEditor).
- Added usage of multilingual input fields in news and events
- Converted the two language fields of content box to one (with custom migration to keep content)
- Moved about.html from the web app to the contentbox app
- Added error message for content boxes, when one or more languages are missing


ddabble:
Should we just leave the admin site for `ContentBox` as it is (just plain text editor for the `content` field and having to manually switch between page languages to edit `content` in another language), or should we try to fix it? Or just leave a note on the admin site to use `/contentbox/<id>/edit` instead..?

ddabble:
Managed to get this working:

MAKE admin contentboxes

The bottom part is with Semantic's CSS, the top part is without. Note that in the bottom one, the whole admin site's CSS is a little messed up, because of Semantic.

zootos:
First of all, I do not see the point of fixing it in the admin page. As far as I know, nobody has ever used the admin page for the content boxes. The only difference between the admin page and the normal editing page, is the title field. Which if changed, will make the contentbox inaccessable due to the urls being defined from this title.

Second, I have tried introducing Semantic UI into the admin panel before. However, the side effects where much greater than the benefits. 

If we were to fix this, the best thing would be to create an extra widget for the field and use this for the admin page.

ddabble:
Sure, just better to not leave the admin site partly unusable, even if it's barely used. The top part of my screenshot does not introduce Semantic UI, and even though it looks a bit shabby, it works. Would that be an acceptable temporary solution?

ddabble:
Also, switching between languages when editing articles and events does nothing :v

zootos:
Created a simple design for the admin panel and created a generic ModelAdmin object, which models with multilingual fields can use when registering for the admin page. A bit unsure about what you mean by switching between languages when editing articles and events does nothing. Are you talking about the admin site?

ddabble:
Nice, will have a look at that :)

I meant switching between languages on the edit page of articles and events, e.g. `/news/article/1/edit`.

ddabble:
Oh whoops, didn't realize that 😅

PR #186, opened 2019-02-03
Tickets

zootos:
Adds support for event registration on the website. This change will replace Hoopla in the long run, however until the current future events have happened, the Hoopla database field will stay (It has been hidden). This change require some changes to the local settings on the server, as it relies on proper configuration of emails.


news/views.py
```diff
@@ -189,12 +203,13 @@ class CreateEventView(PermissionRequiredMixin, CreateView):
         'private',
         'featured',
         'multiday',
-        'hoopla',
```
hermabe: When creating an event the `number_of_tickets` field is missing, but the information box is there. I assume the field should be there?

image

zootos:
> Should the `Single registration` checkbox in the event creation page be moved next to the `number_of_tickets` field, as it enables/disables it?

I tried moving it. However, the flow of the form becomes a bit akward.

zootos:
> For some reason creating and editing of `timeplaces` seem to fail when using the English version of the website. The page just reloads and no `timeplace` is created. The Norwegian version works as expected, creating a `timeplace` and redirecting to the event admin page.
> 
> EDIT: The problem exists on the `master` branch as well.

Seemed to be a date/time format issue. I should be fixed with commit a4260cb.

PR #191, opened 2019-02-10
Misc improvements

ddabble:
* Improved translations
 * Improved ldap_utils documentation
 * Fixed the nav buttons' bubble-backgrounds only partially expanding before the new page loads on fast mobile devices
 * Introduced usage of the `SemanticTimeInput` and `SemanticDateInput` widgets for the `TimePlace` and `Article` creation and edit forms to be able to remove some unnecessary js files
 * Misc code cleanup


make_queue/templates/make_queue/machine_card.html
```diff
@@ -33,7 +33,7 @@
     {% if machine.machine_type.has_stream %}
         <div id="fader"></div>
         <i id="closefullscreen" class="ui times icon"></i>
-        <script>var streamID = "stream-{% sanitize_stream_name machine %}"</script>
+        <script>let streamID = "stream-{% sanitize_stream_name machine %}";</script>
```
zootos: This will be redefined for each stream, so `let` will result in a Javascript error. See the `/reservation/` page as an example.

make_queue/templates/make_queue/js/calendar.html
```diff
@@ -1,8 +1,8 @@
 {% include 'make_queue/js/reservation_rules.html' %}
 <script>
-    var canMakeReservation = {{ can_make_more_reservations|lower }};
-    var canIgnoreRules = {{ can_ignore_rules }};
-    var rules = [
+    let canMakeReservation = {{ can_make_more_reservations|lower }};
+    let canIgnoreRules = {{ can_ignore_rules }};
+    let rules = [
```
zootos: This file is loaded async using ajax on the new reservation page. So all these `let` statements will result in errors each time the week is changed.

make_queue/static/make_queue/js/stream.js
```diff
@@ -30,7 +30,7 @@ $("html").keydown(function (event) {
     }
 });
 
-var closeFullscreen = function () {
+let closeFullscreen = function () {
```
zootos: This will be redefined for each stream, so `let` will result in a Javascript error. See the `/reservation/` page as an example.

make_queue/static/make_queue/js/make_reservation.js
```diff
@@ -1,7 +1,7 @@
 let reservations = [];
 let reservationRules = [];
 let reservationCalendarDate = new Date();
-var canIgnoreRules = false;
+let canIgnoreRules = false;
```
zootos: This `let` will collide with the `canIgnoreRules` variable in the `reservation_calendar.html` file. (Both are used on the new reservation page)

zootos:
In addition to your changes, I introduced the usage of the `SemanticTimeInput` and `SemanticDateInput` widgets for the `TimePlace` and `Article` creation and edit forms to be able to remove some unnecessary js files.

web/static/web/js/widgets/mazemap-search.js
```diff
@@ -40,10 +40,10 @@ jQuery.fn.extend({
             onSelect: function (result, response) {
                 if (data.urlfield) {
                     $(`input[name=${data.urlfield}]`).val(
-                        `https://use.mazemap.com/?campusid=${data.campusid}&desttype=poi&dest=${result.id}`
+                        `https://use.mazemap.com/?campusid=${data.campusid}&desttype=poi&dest=${result.id}`,
```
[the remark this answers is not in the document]
ddabble: It was generated by PyCharm; can remove it!

make_queue/static/make_queue/js/make_reservation.js
ddabble: Hm, it seems to me like this variable is always overwritten by the one in `calendar.html` (assuming that's what you meant with `reservation_calendar.html`, as `calendar.html` is included in it), but correct me if I'm wrong. I think it would be more easily understandable if we either remove it from `make_reservation.js` or write something like `var canIgnoreRules; // Defined in calendar.html`.

zootos: The variable is probably overwritten in all cases, as I think this code was written before the calendar.html code. I like you current solution.

PR #196, opened 2019-02-14
Footer

maiht99:
Added pin footer, updated Slack icon and updated translations.


web/templates/web/footer.html
```diff
@@ -1,6 +1,7 @@
 ﻿{% load staticfiles %}
 {% load i18n %}
 
+<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.7.1/css/all.css"  integrity="sha384-fnmOCqbTlWIlj8LyTjo7mOUStjsKC4pOpQbqyi7RrhN7udi9RwhKkMHpvLbHG9Sr" crossorigin="anonymous">
```
zootos: Could we possibly add just the slack icon. It feels a bit overkill to add `70.9kB` of CSS libraries for a single icon. Especially considering that the library will be loaded on each page.

kalkins: I have modified the FontAwesome font so it only includes the slack icon, which reduced the size to `2.7 kB`

PR #199, opened 2019-02-14
Machine forms

hermabe:
Added functionality to add, edit and delete machines.


make_queue/forms.py
```diff
@@ -153,3 +154,32 @@ class FreeSlotForm(forms.Form):
...
+
+
+class EditMachineForm(BaseMachineForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["status"] = forms.ChoiceField(choices=(
+            ("F", _("Available")),
+            ("O", _("Out of order")),
+            ("M", _("Maintenance")),
+        ))
```
zootos: This does not seem to work

make_queue/forms.py
```diff
@@ -153,3 +154,32 @@ class FreeSlotForm(forms.Form):
...
+        widgets = {
+            "status": SemanticChoiceInput(),
+            "machine_type": SemanticChoiceInput(),
+            "location": MazemapSearchInput(url_field="location_url"),
+        }
+
+
+class CreateMachineForm(BaseMachineForm):
+    class Meta(BaseMachineForm.Meta):
+        exclude = ["status"]
```
zootos: Why would we need to exclude status on creation? I think that it would be more useful to include it, as sometimes (e.g. for the printers) we would like to set it to `Maintenance` or `Out of order` at creation. For sewing machines on the other hand, it would be okay to set it to `Free` at creation.

hermabe: I thought that you wouldn't publish a machine on the website until it was ready. I see the use case now, so can add the field back again.

make_queue/forms.py
hermabe: It works for me. I only get these three options and can create and update machines.
image

What happens at your machine?

zootos: It works now. I was probably just a commit behind or something.

make_queue/forms.py
```diff
@@ -153,3 +154,27 @@ class FreeSlotForm(forms.Form):
...
+            ("O", _("Out of order")),
+            ("M", _("Maintenance")),
+        ))
+
+    class Meta:
+        model = Machine
+        fields = "__all__"
+        widgets = {
+            "status": SemanticChoiceInput(),
+            "machine_type": SemanticChoiceInput(),
```
zootos: Machine type has a widget specified by default.

make_queue/forms.py
```diff
@@ -153,3 +154,27 @@ class FreeSlotForm(forms.Form):
...
+            ("F", _("Available")),
+            ("O", _("Out of order")),
+            ("M", _("Maintenance")),
+        ))
+
+    class Meta:
+        model = Machine
+        fields = "__all__"
+        widgets = {
+            "status": SemanticChoiceInput(),
```
zootos: This does not seem to work for me. I only get the default chrome select, without the Semantic UI classes. It does, however, work if I set the widget in the ChoiceField constructor. I do not think that `widget` in the `Meta` class generally works for self defined fields.

PR #205, opened 2019-03-03
Unique usernames on course registration

hermabe:
Course registration now requires unique usernames.


make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -32,6 +33,16 @@ <h1>
...
                 <p> {% trans "Are you sure you entered the correct username?" %} </p>
             </div>
+            {% form_has_error form "username" "unique" as unique_error %}
+            {% if unique_error %}
+                <div class="ui error message">
+                    <i class="close icon"></i>
+                    <div class="header">
+                        {% trans "Username already exists" %}
+                    </div>
+                    <p>{% trans "Are you sure you entered the correct username?" %}</p>
```
zootos: Should we not have some other message than this in the description part of the unique constraint error. Something along the lines of "A course registration for the given username already exists. Either the user has already taken the course, or the username of another user has been entered."

make_queue/templates/make_queue/course/registration_edit.html
```diff
@@ -32,6 +33,16 @@ <h1>
                 </div>
                 <p> {% trans "Are you sure you entered the correct username?" %} </p>
             </div>
+            {% form_has_error form "username" "unique" as unique_error %}
```
zootos: I think we should use named arguments for the `field` and `code` arguments when calling the template tag. That would make the code more readable. 

```{% form_has_error form field="username" code="unique" as unique_error %}```

File: web/templatetags/util_tags.py [hunk 3]
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
+
+
+@register.simple_tag()
+def form_has_error(form, field, code):
```
zootos: Rename `code` argument to `error_code` for better readability.

make_queue/migrations/0010_auto_20190302_2210.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.1.5 on 2019-03-02 21:10
+
+from django.db import migrations
+import make_queue.models.course
```
zootos: Rename file to `0010_unique_username_in_course_registration.py` to make it easier to see what migrations does (I just recently found out that you could rename the migrations, and think it makes sense to do this for all future migrations).

web/templatetags/util_tags.py
```diff
@@ -0,0 +1,8 @@
+from django import template
+
+register = template.Library()
```
zootos: Rename file to `form_utility.py`?

make_queue/migrations/0010_auto_20190302_2210.py
hermabe: Yes, that is useful.

make_queue/templates/make_queue/course/registration_edit.html
hermabe: Yes, that gives the user more information.

locale/nb/LC_MESSAGES/django.po
```diff
@@ -1452,3 +1459,6 @@ msgstr "Velg verdi"
 #: web/widgets.py:34
 msgid "Search places"
 msgstr "Søk etter steder"
+
+#~ msgid "Username already exists"
```
zootos: Can we delete this unused translation

hermabe: Yes

PR #207, opened 2019-03-04
Hide the apply page

kalkins:
The page is still available through the URL, but is removed from the header.

### DIFF
--- a/web/templates/web/header.html
+++ b/web/templates/web/header.html
@@ -41,10 +41,17 @@
             <div class="text">{% trans "About us" %}</div>
             <div class="make_bg_yellow bubble-background"></div>
         </a>
-        <a class="item" href="{% url 'apply' %}">
-            <div class="text">SØK</div>
-            <div class="make_bg_blue bubble-background"></div>
-        </a>
+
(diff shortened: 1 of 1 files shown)


zootos:
Could it be an idea to add a `Boolean` field to the ContentBox model for cases like this? So we could enable/disable ContentBoxes in the adminpanel, rather than having to modify the code.

kalkins:
@zootos But how would we add/remove the link from the header?

zootos:
There are several ways this could be done. For links on a single page, it is as simple as including the ContentBox object in the context. For the header this is not really an option, as we do not want to remember to include the `apply` ContentBox in all the contexts manually. However, Django allows for extra [global contexts](https://docs.djangoproject.com/en/2.1/ref/templates/api/#django.template.RequestContext). 

A better solution might be to write a simple two line templatetag (e.g. `isContentBoxEnabled(title)`) that returns the value of the `Boolean` field for the ContentBox with the given title.

kalkins:
I've moved the discussion to a new card under ideas on Trello: https://trello.com/c/GlkiRmMv

PR #212, opened 2019-03-16
Internal pages with member list

zootos:
Adds internal pages which are accessible on the subdomains `i`, `internal` and `internt`. The internal pages are quite barebones at the moment, only containing functionality for a member list. In the future we would want to add functionality such as the guidance hours, secrets, "stand lists", useful links for new and old members, the wish list, etc.

Additionally moves the `/admin` page to its own subdomain (`admin.`).

Upgrades Semantic UI from 2.3.0 to 2.4.1

With these changes, running the project locally will change a bit. It is no longer possible to use 127.0.0.1 or localhost as neither of these allow for subdomains. Rather, one has to either bind a random domain to 127.0.0.1 in the local DNS server, or access a domain that has a global DNS entry of 127.0.0.1. For either one of these cases, the properties `SESSION_COOKIE_DOMAIN` and `PARENT_HOST` has to be changed in the settings file. Unless one uses the already configured domain `local.test.pe` running Django on port 8000. For which `local.test.pe:8000` and `<subdomain>.local.test.pe:8000` will work correctly.


kalkins:
I don't like using DNS tricks to be able to run the server locally. Isn't there a way to disable subdomains when DEBUG=True, or something like that?

ddabble:
@kalkins Accessing the local site with `local.test.pe:8000` works :)

Also found out that `localhost:8000` works with subdomains to some degree; i.e. the site can be displayed, but I'm seemingly not able to send data to the local server, like login credentials.

Here's the debug log from (trying to) log in through the admin site, if anyone's interested:
`admin.localhost:8000`:
```
HTTP POST /login/?next=/ 302 [0.47, 127.0.0.1:52660]
HTTP GET / 302 [0.01, 127.0.0.1:52660]
HTTP GET /login/?next=/ 200 [0.04, 127.0.0.1:52660]
```
`admin.local.test.pe:8000`:
```
HTTP POST /login/?next=/ 302 [0.47, 127.0.0.1:52663]
HTTP GET / 200 [0.08, 127.0.0.1:52663]
```
So it seems like the error stems from the server returning 302 from the GET-request..

internal/models.py
```diff
@@ -0,0 +1,174 @@
...
+    date_quit = models.DateField(blank=True, null=True, verbose_name=_("Date quit"))
+    reason_quit = models.TextField(max_length=256, default="", blank=True, verbose_name=_("Reason quit"))
+    comment = models.TextField(max_length=256, default="", blank=True, verbose_name=_("Comment"))
+    active = models.BooleanField(default=True, verbose_name=_("Is active"))
+    guidance_exemption = models.BooleanField(default=False, verbose_name=_("Guidance exemption"))
+    quit = models.BooleanField(default=False, verbose_name=_("Has quit"))
+    retired = models.BooleanField(default=False, verbose_name=_("Retired"))
+    honorary = models.BooleanField(default=False, verbose_name=_("Honorary"))
+
+    def on_create(self):
```
kalkins: I find it slightly messy and hard to read when this is called by an external receiver. You can do the same in save() using `if self.pk == None`

internal/views.py
```diff
@@ -0,0 +1,121 @@
...
+class MemberQuitView(UpdateView):
+    template_name = "internal/member_quit.html"
+    model = Member
+    form_class = MemberQuitForm
+    success_url = reverse_lazy("members")
+
+    def form_valid(self, form):
+        member = form.instance
+        if member.retired or member.quit:
+            raise ValidationError("Bad Request. Tried to set member that is either retired or quit to quit.")
```
kalkins: Slightly confusing error message

zootos:
@kalkins, @ddabble Regarding the subdomains locally, I am looking into a fix where I will bind the views of the subdomains to `/subdomain/<subdomain_name>/`. I am having some problems with i18n patterns and reversing, however nothing that cannot be solved.

@ddabble `localhost` won't remember your login data because the session cookie is set to bind to `local.test.pe`

internal/models.py
zootos: Yeah, the receiver logic for creation is remnant from when updates to the committées went through it. I agree that an override in `save()` is better here now.

internal/views.py
zootos: Yeah, will have a look at it.

ddabble:
@zootos Oh okay, I see!

The subdomain fix sounds great, but is that necessary, considering `local.test.pe` works just fine? Also, would `/subdomain/<subdomain_name>/` only work i debug mode? Because if so, I'd personally prefer if the site paths were the same for both the local and production server, apart from the domain name.

Is it possible to bind the session cookie to `localhost` instead?

zootos:
@ddabble It is possible to bind it to localhost instead. I just did not realise that you could actually use subdomains directly on localhost (e.g. `admin.localhost`).

@kalkins If bind it to localhost instead, is that fine by you? Then you should not have to perform any DNS tricks. This works directly out of the box for Chrome and Opera, but for Firefox you have to go to 
`about:config` and set `network.dns.native-is-localhost` to `true`.

kalkins:
@zootos Yes, that's better

zootos:
Okay, then I will drop the `/subdomain/<subdomain_name>/` fix.

zootos:
One problem, binding the session cookie to `localhost` is not possible, as we want to have the same session across subdomains. And domains for cookies have to have at least 2 dots in them, or be empty (empty means only this specific subdomain). So for `localhost` to work, we would have to bind the page to a subdomain of localhost, e.g. `make.localhost`. Then everything can be accessed on the domains: `admin.make.localhost`, `(i|internal|internt).make.localhost` and `make.localhost`.

kalkins:
That works great, and it also solved some port confusion I've been having locally

zootos:
@kalkins @ddabble The page is now bound to `makentnu.localhost`. This actually makes page accesses look quit similar to the main web page (`admin.makentnu.localhost` vs. `admin.makentnu.no`). Also, I have fixed the issues that @kalkins pointed out earlier, so the PR should be ready to be reviewed.

kalkins:
The internal/home page is very empty. Maybe redirect to internal/members until we have something to put there?

zootos:
> The internal/home page is very empty. Maybe redirect to internal/members until we have something to put there?

I agree. The internal/home page now just shows the members list view.

PR #214, opened 2019-03-21
Disable Dataporten login if not configured

zootos:
Disables login through Dataporten if Dataporten is not configured (i.e. secret is empty). This actually makes the login button work when running locally.

Disables login through the Django admin panel login view if Dataporten is configured. Not really needed, however, we only really want logins through the Dataporten in production anyways. And if there for some odd reason is a vulnerability in this page at some point, we do not need to care about it. Also removes the risk of somebody bruteforcing the password of a superuser (if we by mistake set a password for a superuser in production at some point).


zootos:
Also added support for redirecting to subdomains. That means that accessing `internal.makentnu.no` without actually being logged in, will redirect to `internal.makentnu.no` after login.

web/urls.py
```diff
@@ -45,6 +40,25 @@
...
+if settings.SOCIAL_AUTH_DATAPORTEN_SECRET:
+    urlpatterns += i18n_patterns(
+        path('login/', RedirectView.as_view(url='/login/dataporten/', query_string=True), name='login'),
+        path('logout/', Logout.as_view(), name='logout'),
+        re_path(r'^complete/(?P<backend>[^/]+){0}$'.format(extra), login_wrapper),
+        path('', include('social_django.urls', namespace='social')),
+        prefix_default_language=False,
+    )
+
+# If it is not configured, we would like to have a simple login page. So that
```
kalkins: I would prefer if this comment was inside the else block. Now the if and else look detached

kalkins:
Is ldap required for anything other than login? We have had some problems installing it on windows, so it would be great if we didn't have to

zootos:
> Is ldap required for anything other than login? We have had some problems installing it on windows, so it would be great if we didn't have to

LDAP is used for two different cases:
- Dataporten login to find basic user data (full name, and NTNU username)
- For course registration: Checking if a given username is valid, and in that case retrieve the users full name 

@hermabe is the one that have done most of the implementation work, and he is running Windows 10. So he probably have some idea about what you need to do to get it working.

web/urls.py
zootos: I agree that it becomes more readable, have moved it now.

hermabe:
I run the website using Linux subsystem for Windows.

Installing `python_ldap` with pip does not seem to work. However https://python-ldap.readthedocs.io/en/latest/installing.html gives some instructions. I downloaded one of the pre-built binaries and it seemed to work fine.

kalkins:
Sorry, I didn’t explain properly. It’s not impossible to install ldap, we’ve made it work for everyone in dev who uses windows. However, it’s a pain to install, and the first few times it took forever to find out how to do it right. Even on linux it’s slightly annoying having to install a system package outside pip.

Ideally we should only have to do `pip install`, and everything should work. My point was that if ldap isn’t required for local use, then everyone shouldn’t have to install it.

zootos:
Did not realise that it was that problematic to install on other OSes. On MacOS it works right out of the box with pip. However, since it is used in the course registration, we probably cannot easily disable it locally.

PR #218, opened 2019-03-24
Misc improvements

ddabble:
* Fixed offset checkbox click area on the edit member page
 * Added formatting of phone numbers
 * Added a couple missing translations
 * A couple accessibility improvements
 * Code cleanup


internal/models.py
```diff
@@ -52,6 +52,14 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
             # Add user to the MAKE group
             self.toggle_membership(True)
 
+    @property
+    def formatted_phone_number(self):
+        if len(self.phone_number) == 8:
+            # Safe to assume that the phone number is Norwegian
+            return "{} {} {}".format(self.phone_number[:3], self.phone_number[3:5], self.phone_number[5:])
```
zootos:
```python
return f"{self.phone_number[:3]} {self.phone_number[3:5]} {self.phone_number[5:]}"
```
is more readable in my opinion.

internal/forms.py
```diff
@@ -39,6 +39,11 @@ def __init__(self, user, **kwargs):
             for field_name in ["committees", "role", "comment", "guidance_exemption", "active", "honorary"]:
                 del self.fields[field_name]
 
+    def clean_phone_number(self):
+        phone_number = self.cleaned_data['phone_number']
+        # Remove spaces from phone number, as it'll be formatted (by `Member.formatted_phone_number()`) when displayed
+        return phone_number.replace(" ", "")
```
zootos: Is there are a reason why we do not do this in one operation?
```python
return self.cleaned_data['phone_number'].replace(" ", "")
```

Also, for our use case we do not really want anything besides numbers (maybe a +). So we could consider using something along the lines of:
```python
return re.sub("[^\d]", "", self.cleaned_data["phone_number"])
```

Additionally, depending on the solution we choose, it might be smart to add some Javascript to the frontend to reflect these changes. E.g. code that removes all characters not numbers (we cannot use `input[type=number]`, since this allows for the letter e, and does not allow for leading zeros). Also, it might be an idea to add some more styling to the input field: 
Screenshot 2019-03-25 at 11 49 53

kalkins: There is a django plugin for a phone number model field which handles validation and printing quite well: https://github.com/stefanfoulis/django-phonenumber-field

kalkins: I've incorporated it here: https://github.com/MAKENTNU/web/pull/222

PR #221, opened 2019-03-25
Misc fixes

zootos:
Contains a variety of small quality of life changes:
- Throws 404 instead of showing an empty page, when accessing `/admin` without any of the permissions required.
- `robots.txt` files for `admin.` and `(internal|i|internt).` that disallows any crawling of the subdomains
- Migrate machine usage rules to the "new" multilingual ckeditor fields. This also fixes a bug, where the  usage rules where un-editable.
- Adds a hover text for the mail icon in the footer
- Fixes some weird formatting of the social media icons in the footer when the screen is between 900 and 1000 pixels wide
- Adding a member to the members list now requires a user to be selected.


web/views.py
```diff
@@ -22,6 +23,20 @@ def get_context_data(self, **kwargs):
...
+        "news.add_article", "news.change_article", "news.delete_article",
+        "news.add_event", "news.change_event", "news.delete_event",
+        "news.add_timeplace", "news.change_timeplace", "news.delete_timeplace",
+        "make_queue.can_create_event_reservation",
+        "make_queue.change_quota",
+        "make_queue.change_printer3dcourse",
+        "groups.can_edit_group",
+    ]
+
+    def get_context_data(self, **kwargs):
```
kalkins: Shouldn't this be done with a `PermissionRequiredMixin` or `UserPassesTestMixin`?

kalkins: Also, this is technically a 403 error

zootos: `PermissionRequiredMixin` does not work in this situation, as somebody should have access to the admin panel even though they do not have access to all the admin subsystems. Never seen `UserPassesTestMixin` before, but seems to be a good fit. Also totally agree that it should be 403.

PR #222, opened 2019-03-28
Add django-phonenumber-field

kalkins:


internal/migrations/0004_auto_20190328_1736.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.1.7 on 2019-03-28 16:36
+
+from django.db import migrations
```
zootos: Could we rename this migration, so to make it easier to see what was changed in each migration? E.g. `0004_member_phone_number_field`

kalkins: Sure

zootos:
Should probably add an error message on submit with an invalid phone number, as currently the form fails to submit with no error message. This makes it almost impossible for the user to know what is wrong.

PR #227, opened 2019-04-15
Upgrade to Django 2.2

zootos:
Fixes failing tests due to Django automatically upgrading to version 2.2.0. We do not want this yet, as we have not been able to test if the website works for this version.


zootos:
Decided to update to Django 2.2 instead of keeping 2.1, as there were only small changes required for 2.2.

PR #228, opened 2019-04-16
Redirect the user to their current page on login

zootos:
Redirects the user to their current page on login. Is dependent on #227

### DIFF
--- a/web/context_processors.py
+++ b/web/context_processors.py
@@ -1,0 +1,23 @@
+from django.contrib.auth.views import LoginView
+from django.shortcuts import reverse
+
+
+def login(request):
+    # The current `next` parameter
+    current_redirect_url = LoginView(request=request).get_redirect_url()
+
+    # If at the login page:
+    if request.path.startswith(reverse('login').rstrip("/")):
+        # Let the `next` parameter be the same as it currently is
+        login_redirect_path = current_redirect_url
+    # If at the front page:
+    elif request.path == "/":
+        # Don't redirect after login
+        login_redirect_path = None
+    else:
+        # Redirect back to the same page after logging in
+        login_redirect_path = request.path
+
+    return {
+        'login_next_param': ("?next=" + login_redirect_path) if login_redirect_path else "",
+    }
--- a/web/context_processors.py
+++ b/web/context_processors.py
@@ -19,5 +19,5 @@ def login(request):
         login_redirect_path = request.path
 
     return {
-        'login_next_param': ("?next=" + login_redirect_path) if login_redirect_path else "",
+        'login_next_param': f"?next={login_redirect_path}" if login_redirect_path else "",
     }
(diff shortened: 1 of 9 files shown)


zootos:
@ddabble Is there a reason for b340c57. The front page redirect works by the same logic as all the other pages. While the front don't need the `next` parameter, I think the change of b340c57 just increases the complexity of the code, without actually adding any value.

ddabble:
@zootos The only reason for it is stated in the commit message:
> This makes no functional difference, but leaves the login URL clean when possible.

So I like it because it makes the most used login URL (assuming most people log in from the front page) more clean-looking, but this is pretty trivial and might as well be removed.

zootos:
@ddabble I got that. My point was more a question about if this single clean URL (that most people probably never look at) is worth the extra code complexity it adds to the template.

ddabble:
@zootos Was unsure about if it was worth the extra complexity or not.. :p Anyway, I moved the logic to a context processor and at the same time fixed some (probably) unintended behavior.

web/context_processors.py
```diff
@@ -0,0 +1,23 @@
+from django.contrib.auth.views import LoginView
+from django.shortcuts import reverse
+
+
+def login(request):
+    # The current `next` parameter
+    current_redirect_url = LoginView(request=request).get_redirect_url()
+
+    # If at the login page:
+    if request.path.startswith(reverse('login').rstrip("/")):
```
zootos: Is there a reason why some strings use single quotes?

web/context_processors.py
```diff
@@ -0,0 +1,23 @@
...
+    # If at the front page:
+    elif request.path == "/":
+        # Don't redirect after login
+        login_redirect_path = None
+    else:
+        # Redirect back to the same page after logging in
+        login_redirect_path = request.path
+
+    return {
+        'login_next_param': ("?next=" + login_redirect_path) if login_redirect_path else "",
```
zootos: Here it might be more readable to use f-strings, rather than string concatenation.

web/context_processors.py
ddabble: I'm used to referring to specific defined values (practically the same as enums) with single quotes and everything else with double-quotes.

To take this file as an example, `'login'` is the defined name of a URL, `"/"` is just a path (one could argue that it's borderline a defined value), `'login_next_param'` will become a defined value that's used in the template, and `"?next="` is a constructed string when combined with `login_redirect_path` (although I guess just `'next'` would be a defined value).

Ultimately, it's only for giving hints as to which values are defined elsewhere in the code, but otherwise not really important.

web/context_processors.py
ddabble: You mean like this?
```python
'login_next_param': f"?next={login_redirect_path}" if login_redirect_path else ""
```

PR #229, opened 2019-05-04
Remove SiT-logo from footer

shahzebs:


ddabble:
Why are we removing the SIT-logo, btw?

zootos:
> Why are we removing the SIT-logo, btw?

The period for our sponsorship with SIT has ran out. Thus, we are no longer required to have their logo on the website, and we should ideally one have logos on our website if we are receiving some benefit from it.

ddabble:
Ah, I see 👍

PR #238, opened 2019-09-26
Card

hermabe:
Created a card app for saving all information about card numbers, connected to the user. Previously the information was saved three places: `internal.Member`, `make_queue.Printer3DCourse` and `checkin.Profile`. 

By doing this, there is no need to enter information already entered other places. For example, when adding a new member and the member have been to the printing course, the card number will already be connected to the user so it will be shown in the member list.

I may have missed some cases, and there is probably room for improvement.


hermabe:
> The migrations fail when you try to apply them to the production database

I have now removed the duplicates.

zootos:
Migration of prod data still crashes. It finishes the `card.0001_initial` and `checkin.0007_remove_profile_card_id` migrations fine, and then it crashes on the `internal.0005_remove_member_card_number` migration because of duplicate card numbers. I think this is because you don't use the `update_or_create` function in the migrations. So the members who have taken the 3D-printer course will result in duplicate card numbers.

card/models.py
```diff
@@ -0,0 +1,33 @@
...
+class Card(models.Model):
+    """
+    Model for connecting a card number to a user
+    """
+    number = models.fields.CharField(max_length=16, verbose_name=_("Card number"))
+    user = models.OneToOneField(User, on_delete=models.CASCADE, verbose_name=_("User"))
+
+    class Meta:
+        constraints = (
+            UniqueConstraint(fields=['number'], name='unique_number'),
```
zootos: This way of creating unique constraints is meant for more complicated functionality. For just making sure that a single field has unique values, just use `unique=True` when declaring the field. It also makes the code more readable.

zootos: I think it would, however, be useful to create custom constraints for what values the number field can take. Such as being of a certain length and just being numbers.

checkin/views.py
```diff
@@ -143,6 +95,12 @@ def get_context_data(self, **kwargs):
             profile = Profile.objects.get(user=self.request.user)
         else:
             profile = Profile.objects.create(user=self.request.user)
+
+        # Connect card number from course registration to user
+        registration = Printer3DCourse.objects.filter(user__username=self.request.user.username)
+        if registration.exists():
+            Card.update_or_create(self.request.user, "EM " + str(registration.first().card_number))
```
zootos: I think we should not add "EM " to all the card numbers. It is much better if we just prepend it in the interface when we feel the need to show it. Until now, we have only collected the numbers and not prepended them with "EM".

internal/forms.py
```diff
@@ -35,6 +37,11 @@ class Meta:
 
     def __init__(self, user, **kwargs):
         super().__init__(**kwargs)
+
+        member = kwargs['instance']
+        if hasattr(member.user, 'card'):
+            self.initial['card_number'] = member.user.card.number.split()[1]  # Without EM prefix
```
zootos: Here we again have the extra problems that prepending "EM" to the card numbers give.

internal/forms.py
```diff
@@ -23,6 +23,8 @@ def __init__(self, *args, **kwargs):
 
 
 class EditMemberForm(ModelForm):
+    card_number = IntegerField(label=_("Card number (EM)"))
```
zootos: Maybe we should make a custom field and widget for EM numbers, that perform validation and have [a label](https://fomantic-ui.com/elements/input.html#labeled) on the input field that says EM.

internal/views.py
```diff
@@ -37,6 +39,14 @@ class AddMemberView(PermissionRequiredMixin, CreateView):
     def get_success_url(self):
         return reverse_lazy("edit-member", args=(self.object.pk,))
 
+    def form_valid(self, form):
+        # Connect card number from course registration to user
+        user = form.cleaned_data['user']
+        registration = Printer3DCourse.objects.filter(username=user.username)
+        if registration.exists():
+            Card.update_or_create(user, "EM " + str(registration.first().card_number))
```
zootos: Again the thing further up with prepending "EM"

internal/views.py
```diff
@@ -51,6 +61,13 @@ def get_form(self, form_class=None):
             form_class = self.get_form_class()
         return form_class(self.request.user, **self.get_form_kwargs())
 
+    def form_valid(self, form):
+        # Update card connected to user
+        user = self.object.user
+        card_id = form.cleaned_data["card_number"]
+        Card.update_or_create(user, "EM " + str(card_id))
```
zootos: Again the thing further up with prepending "EM"

make_queue/fields.py
```diff
@@ -29,6 +30,8 @@ def can_use_3d_printer(user):
         course_registration = Printer3DCourse.objects.get(username=user.username)
         course_registration.user = user
         course_registration.save()
+        if course_registration.card_number:
+            Card.update_or_create(user, "EM " + str(course_registration.card_number))
```
zootos: Again the thing further up with prepending "EM"

make_queue/forms.py
```diff
@@ -147,6 +148,11 @@ class Meta:
             "username": forms.TextInput(attrs={"autofocus": "autofocus"}),
         }
 
+    def save(self, commit=True):
+        if self.cleaned_data['user'] and self.cleaned_data['card_number']:
+            Card.update_or_create(self.cleaned_data['user'], "EM " + str(self.cleaned_data['card_number']))
```
zootos: Again the thing further up with prepending "EM"

zootos:
Seems like there wasn't errors in the migration code, but duplicates in the production data.

card/models.py
ddabble: Regarding constraining the `number` field to only contain numbers: we could try using the [int-list-validator](https://docs.djangoproject.com/en/2.2/ref/validators/#int-list-validator) with `sep=''`..?

zootos: I havn't looked at this for too long, but `int-list-validator` seems a bit complex. Also, just on a whim I think it might be good to have a more strict constraint. Such as numbers of X or Y digits. This, as we know that EM numbers have to be of a specific length (or two specific sizes when we also consider the old numbers).

hermabe: Old numbers can be zero-padded to get 10 digit numbers for them as well. The numeric value works the same way as the new ones and can be read by the card reader in the same way. Something like `\d{10}´ could work.

hermabe:
Couple changes:
 - Removed the "EM" prefixes
 - Added custom model and form field for card numbers

With the extra validation not all `Printer3DCourse` card numbers in the production database will pass as they are too long.

card/models.py
hermabe: Currently implemented the validation with a `RegexValidator` with the pattern `^\d{10}$`.

card/views.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Base view class for receiving requests from RFID card readers
+    """
+
+    @method_decorator(csrf_exempt)
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+    @csrf_exempt
+    def post(self, request):
```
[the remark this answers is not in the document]
hermabe: According to [this example](https://docs.djangoproject.com/en/2.2/ref/csrf/#view-needs-protection-for-one-path) that seems to be correct.

card/models.py
```diff
@@ -0,0 +1,70 @@
...
+        if "form_class" not in kwargs:
+            kwargs["form_class"] = card.forms.CardNumberField
+        return super().formfield(**kwargs)
+
+
+class Card(models.Model):
+    """
+    Model for connecting a card number to a user
+    """
+    number = CardNumberField(unique=True)
```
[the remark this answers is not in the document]
hermabe: `CardNumberField` has "Card number" as a default `verbose_name`, so it is not necessary to set it here.

card/admin.py
```diff
@@ -0,0 +1,11 @@
+from django.contrib import admin
+
+from card.models import Card
+
+
+class CardAdmin(admin.ModelAdmin):
+    model = Card
+    list_display = ('number', 'user',)
```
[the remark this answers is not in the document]
hermabe: Every other display prepends "EM", so I think we should be consistent.

card/models.py
```diff
@@ -0,0 +1,70 @@
...
+    user = models.OneToOneField(User, on_delete=models.CASCADE, verbose_name=_("User"))
+
+    def __str__(self):
+        return "EM " + str(self.number)
```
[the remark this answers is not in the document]
hermabe: Yes, the `str` is a remnant from when the card number was saved as an integer.

card/templates/card/widgets/card_number_input.html
```diff
@@ -0,0 +1,25 @@
...
+
+<script>
+    let required = $("input[name={{ widget.name }}]").prop("required");
+    let pattern = required ? /^[0-9]{7,10}$/ : /^([0-9]{7,10})?$/; // Allow input to be empty if not required
+    let cardNumberLengthCount = $("#card_number_count");
+    $("input[name={{ widget.name }}]").on("keyup paste", (event) => {
+        cardNumberLengthCount.text(event.target.value.length);
+        if (!(pattern.test(event.target.value))) {
+            $("#card_number_error").removeClass("hidden");
+        } else {
```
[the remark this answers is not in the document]
hermabe: Yes, that is easier.

card/admin.py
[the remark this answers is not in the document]
zootos: Yeah, that was my point with `__str__`, as this would prepend "EM", while `number` does not

card/models.py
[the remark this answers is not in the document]
zootos: My bad, didn't see that

hermabe:
I think your solution with a `OneToOneField` from `Printer3DCourse` to `Card` will be the easiest. 

To enforce the users to be equal I think we can add something like this to `Printer3DCourse`. This doesn't solve the problem, but it will keep the database in a consistent state.
```
class Meta:
    constraints = [
        models.CheckConstraint(check=Q(card__user=user), name='user_equal'),
        models.CheckConstraint(check=Q(user__isnull=True) | Q(user__username=username), name='username_equal'),
    ]
```

zootos:
Sounds good

hermabe:
Turns out Django doesn't allow cross table constraints. I get `FieldError: Joined field references are not permitted in this query` when running the migrations.

I don't know if there is another way to enforce these constraints.

ddabble:
Some thoughts on doing validation across tables/models:

We could override the model's `save()` method and do the validation there; a downside to this is that it's usually not expected having to catch a `ValidationError` when saving an object; also `save()` is not called if `update()` is called on the model..

A second option is to only do the validation in forms (through either the model's or the form's `clean` methods), but it might be challenging getting everyone to remember to not call `save()` without validating first.

A third option is to allow the models to contain invalid data, but send an email to the devs when it's registered - e.g. through doing the validation in a `pre_save` or `post_save` signal. This should probably also be considered if we're going for the second option.

hermabe:
I am currently writing a `save()` method for `Printer3DCourse` that in theory should not commit any invalid data to the database. However as you say, `save()` is not called on `update()` or other bulk operations.

[Composite Foriegnkeys](https://sqlite.org/foreignkeys.html#fk_composite) may solve the database constraint problem and are supported by Sqlite, but not Django. We thus have to write the SQL for the migrations manually. Sqlite also does not support adding constraints to an already existing table, so one would have to create a new table with the constraints and copy all the data. (This is how Django does normal migrations as well.) The raw SQL must be tested properly so we dont lose any data.

hermabe:
I have implemented the syncronization by overriding the `save()` methods on `Card` and `Printer3DCourse`. I have not tested all cases yet, but the general solution should work.

ddabble:
> [Composite Foriegnkeys](https://sqlite.org/foreignkeys.html#fk_composite) may solve the database constraint problem and are supported by Sqlite, but not Django. [...]

Hm, interesting! Seems like PostgreSQL supports composite foreign keys as well (in case we would switch to Postgres, which we've discussed briefly before) 🙂

zootos:
I don't mind switching to PostgreSQL, but I think we should try to reconsider if there is any other way to do this. E.g. redesigning the models.

hermabe:
Changing the user model turned out to be much easier than thought. This simplifies a the problem quite a bit. There is no longer use for the `Card` model.

hermabe:
When running migrations on a non-empty database they will fail. Instead of running the `0001_initial` migration for the users app, one must run the following SQL in `python manage.py dbshell`

hermabe:
There has been a lot of back and forth so please look for any remnants from earlier solutions.

hermabe:
This has become fairly messy. I will clean up the commits and create a new PR.

Closing this.

hermabe:
This has become fairly messy. I will clean up the commits and create a new PR.

Closing this one.

PR #240, opened 2019-10-24
Custom user model

hermabe:
Changing the default Django user model for a custom one, so that we can add extra data to users in a simpler way.  Currently there is no extra data added, so no functional change.

The users app `0001_initial` migration will fail on a non-empty database. Instead of running the `0001_initial` migration one must run the following SQL in `python manage.py dbshell`

```
INSERT INTO django_migrations (app, name, applied) VALUES ('users', '0001_initial', CURRENT_TIMESTAMP);
UPDATE django_content_type SET app_label = 'users' WHERE app_label = 'auth' and model = 'user';
```

On an empty database the migrations run as normal.


users/migrations/0001_initial.py
```diff
@@ -0,0 +1,52 @@
...
+import django.contrib.auth.validators
+from django.db import migrations, models
+import django.utils.timezone
+
+"""
+This migration will fail if the database is non-empty. One has to manually replace the default Django user model
+with the new one in the database. This is done by running the following SQL in the django dbshell:
+
+INSERT INTO django_migrations (app, name, applied) VALUES ('users', '0001_initial', CURRENT_TIMESTAMP);
+UPDATE django_content_type SET app_label = 'users' WHERE app_label = 'auth' and model = 'user';
```
kalkins: I understand why this is necessary, but I'm reluctant to apply this to our main database just like that. Where did you get this SQL? It isn't listed in the django documentation. Have you tested it thoroughly?

hermabe: I followed [this tutorial](https://www.caktusgroup.com/blog/2019/04/26/how-switch-custom-django-user-model-mid-project/), and it is where I found the SQL.

I think @zootos tested the migrations on the prod database. But I agree that more testing can be done.

zootos: It is currently running on the dev environment with a copy of the prod database. Everything seems to be working fine. Feide login works and all tables with references directly to users point to the new user table. That is, all references to users still work.

zootos: @kalkins Any opinions on this?

kalkins: Sorry, I've been busy for a while and forgot about this. I hate that we have to do it this way, but if it really is the only way, and you've tested that it works, then I'll approve it

ddabble:
@hermabe What's the [`dump.rdb`](https://github.com/MAKENTNU/web/blob/55db24fccb99f33692e2876ae3b33a0e1a756cce/dump.rdb) file for..? 😮

hermabe:
@ddabble That one shouldnt be there.

PR #241, opened 2019-10-24
Card number field on user model

hermabe:
Replacement for #238 . Cleaned up the commits a bit. 

Adds a card number field to the user model. This makes the use of card numbers much simpler and replaces the three different places where they were stored previously.

Depends on #240 , wait for that one to be merged.


kalkins:
I think this didn't transfer my card number correctly. I'll look more into it later

hermabe:
Now that #240 is merged. This is ready to be reviewed.

zootos:
Migration 0013 for `make_queue` fails on the `user_or_cardnumber_null` constraint. I think it might be that the migration does not check if there is a user connected to the course registration. I think you might need to change the order of the migrations so that migration 0003 of `users` happens before migration 0013 of `make_queue`.

hermabe:
I will look over the migrations and test them some more.

hermabe:
Reordered the migrations as @zootos suggested. Now the conversion function ensures that the constraint is met.

Also added some checks so that a null value should never be written thus deleting data. This may have been the issue @kalkins mentioned earlier.

kalkins:
Ok, it looks good. However I have some questions about how registrations and users are handled, as I haven't used that system much, and I don't have time to dig through the whole code:
- As I understand it, we can register that a user has taken a 3D-printer course without them being a user on the site (before they've logged in with Feide). If they log in after that, is the card number on the registration copied to the user model?
- I noticed that some people have completed the course, but we don't have a card number registered for them. Is this a bug, or has the number been lost somehow?

hermabe:
- Yes, the user is connected to the course registration when checking wether they can use the 3d-printer. The card number is then copied to the user model and removed from the course registration.

- Not all users have cards if they are new to NTNU, some forgot to bring them to the printer course. Also, there was a long period in 2017 where card numbers were not registered.

PR #243, opened 2019-10-28
Remove out of order and maintenance machines from find reservations list

sigtheidiot:


ddabble:
@maiht99 Why was this closed? 😮

make_queue/views/reservation/reservation.py
```diff
@@ -298,7 +298,8 @@ def form_valid(self, form):
 
         periods = []
         for machine in Machine.objects.filter(machine_type=form.cleaned_data["machine_type"]):
-            periods += self.get_periods(machine, required_time)
+            if not machine.get_status() in "OM":
```
hermabe: I dont think we should exclude machines under maintenance. Maintainance is short-term and the machine is expected to be operational soon. If a machine is out of order it may be long-term, so it makes sense to exclude them.

kalkins:
Maybe this should be an explicit attribute on the printers, instead of being inferred from other attributes?

zootos:
> Maybe this should be an explicit attribute on the printers, instead of being inferred from other attributes?

Do you mean the status?

kalkins:
I meant that I can imagine that there will be functioning printers we want to hide (like when we lend them out) and printers under maintenance we want to either show or hide, depending on how long the repairs take. So, whether it is shown or not should not be determined by the status, but by a separate attribute that can be toggled in the admin panel.

kalkins:
@hermabe @zootos What do you think?

make_queue/views/reservation/reservation.py
sigtheidiot: Yes, good point. I'll include the machines under maintenance again

zootos:
> @hermabe @zootos What do you think?

I would say that we don't necessarily need to address that problem in this PR. It is an important point to address, but it seems a bit out of scope for this PR. Nevertheless, I can mention that I think maintenance should be a reservation type rather than a state of the machine. Since, as @hermabe points out, maintenance is temporary.

kalkins:
Okay, so then this PR will stay as it is (except not hiding machines under maintenance), and we'll add a card on Trello to move maintenance to a reservation instead of a state. Then we'll see how we hide them after that.

kalkins:
@sigtheidiot You just made two commits with the same title, and with a word misspelled. I won't bother to fix this now, but in the future please look through the log before pushing to make sure it looks nice

hermabe:
I agree with making maintenance a reservation type. This will also make it possible to schedule maintenance, possibly recurring.

PR #244, opened 2019-11-06
Equipment list

siireen:
Template for makerspace and tools functionality.
Admin panel for tools and makerspace.
Added translation for tools and makerspace in admin panel.


requirements.txt
```diff
@@ -9,7 +9,7 @@ channels
 channels_redis
 django_ical
 xlsxwriter
-python_ldap==3.1.0
+python_ldap==3.2.0
```
hermabe: Why change this? What does this have to do with `tools`?

Also, any version `>=3.0` supports Python 3 and should work.

ddabble: It probably doesn't have anything to do with `tools`, but it's a welcome fix nonetheless, because `v3.1.0` isn't available anymore from the [officially linked](https://www.python-ldap.org/en/latest/installing.html#windows) [site](https://www.lfd.uci.edu/~gohlke/pythonlibs/#python-ldap) storing the pre-built binaries for Windows; `v3.2.0` is, though 🙂

makerspace/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.urls import path, register_converter
+
+from makerspace import views
+
+urlpatterns = [
```
zootos: Raw strings are not needed as they only escape backslashes, which should in general never be used in `urls`. It is true that previously, Django used regexes for their urls, but unless you use `re_path` this is no longer the case. So using raw strings gives no additional gain and are in general not used for the `path` function.

makerspace/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class ViewMakerspaceView(TemplateView):
+    template_name = 'makerspace/makerspace.html'
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['latest_makerspace_announce'] = Makerspace.objects.last()
+        print("QP", context)
+        print(context['view'])
```
zootos: These prints should not be here

makerspace/views.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class ViewAdminMakerspaceView(PermissionRequiredMixin, UpdateView):
+    model = Makerspace
+
+    template_name = 'makerspace/admin_makerspace.html'
+    context_object_name = 'makerspace'
+    permission_required = 'makerspace.add_Makerspace'
+    fields = (
+        'title',
```
zootos: If we are using all fields, use `fields = "__all__"` instead. This goes for the other views below as well.

makerspace/models.py
```diff
@@ -0,0 +1,29 @@
+from django.db import models
+
+from web.multilingual.database import MultiLingualTextField, MultiLingualRichTextUploadingField
+
+
+class Tool(models.Model):
+    title = MultiLingualTextField(
+        max_length=100,
+        verbose_name=('Title'),
```
hermabe: The `verbose_name` fields should maybe be translated. This can be done using pythons [gettext_lazy](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.translation.gettext_lazy).

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,90 @@
+{% extends "web/base.html" %}
+{% load staticfiles %}
+{% load i18n %}
+{% load news_tags %}
+
+{% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "news/css/admin.css" %}">
```
hermabe: Why is the news css used?

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,90 @@
...
+{% endblock head %}
+
+{% block body %}
+    <h1 class="ui text-centered make_col_yellow header">
+        {% trans "Tools" %}
+        <a class="make_col_blue explanation-popup" data-title="{% trans "Create" %}"
+           data-content="{% trans "If clicked, a new Tool will be created." %}" href="{% url 'makerspace-tools-create' %}">
```
hermabe: Maybe replace the tooltip with something like "Create a new tool"? It is a button, the clicking part is implicit.

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,90 @@
...
+                            <div class="header">
+                                <a class="inline-block make_col_blue header" href="{% url 'makerspace-tools-edit' tool.id %}">
+                                    {{ tool.title }}
+
+
+                                </a>
+                                {% if perms.tool.delete_event %}
+                                    <a class="delete confirm inline-block make_col_blue explanation-popup"
+                                       data-title="{% trans "Delete" %}"
+                                       data-content="{% trans "If clicked, the tool will be deleted." %}"
```
hermabe: Same as the other "If clicked"

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,90 @@
...
+           data-content="{% trans "If clicked, a new Tool will be created." %}" href="{% url 'makerspace-tools-create' %}">
+            <i class="ui add icon"></i>
+        </a>
+    </h1>
+    <div class="ui text container">
+        {% if tools_list %}
+            <div class="ui divided list">
+                {% for tool in tools_list %}
+                    <div class="item">
+                        <i class="ui icon"><img class="inline square_36" src="{{ tool.image.url }}"></i>
```
hermabe: When I created a tool without an image and was redirected to the tools admin page this fails. I get an `ValueError: The 'image' attribute has no file associated with it.`.  Either check if a tool has an image before trying to show it, or require all tools to have an image in the tool form.

requirements.txt
kalkins: Should probably be a separate PR though

makerspace/converters.py
```diff
@@ -0,0 +1,13 @@
+from django.shortcuts import get_object_or_404
+
+from makerspace.models import Makerspace
+
+
+class ToolbyTitleConverter:
```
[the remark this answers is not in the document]
kalkins: Also, if it's going to be used it would make it easier to read by capitalizing the B

makerspace/templates/makerspace/makerspace.html
```diff
@@ -0,0 +1,43 @@
...
+    <div class="make_content_padding"></div>
+    <div class="ui text container" style="position: relative; margin: 100px 0 ;">
+        {% if user.is_superuser %}
+            <p></p>
+            <a href= "{{ latest_makerspace_announce.id}}/edit"    class="ui yellow button">
+               Rediger
+            </a>
+        {% endif %}
+        <h1>
+            <strong>
```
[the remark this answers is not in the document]
kalkins: Use `font-weight` in the css instead

makerspace/models.py
kalkins: They should, the translations are used in the automatic labels in forms. We usually import `gettext_lazy as _`, so we write `verbose_name=_('Title')`

makerspace/templates/tools/admin_tool.html
kalkins: If the same rules are used in both tools and news, you probably want to create a separate file with the common rules under `web/static/`, and specific rules in each app. If not changes in news could cause problems with tools later

requirements.txt
ddabble: True, but as it's already part of the PR and it's such a small change, I would personally not bother removing it.

zootos: I agree that a separate PR is better. It would have been okay if it was only part of this one, but it is also a separate change for the same thing in #245. So I just made #247 for simplicity.

makerspace/models.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    image = models.ImageField(verbose_name=('Image'), blank=True, )
+    content = MultiLingualRichTextUploadingField()
+
+
+class Makerspace(models.Model):
```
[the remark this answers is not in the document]
qpb96: Comments have been added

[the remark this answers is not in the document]
zootos: What do you mean by add description?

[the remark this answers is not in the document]
qpb96: The intention of having two different classes is to distinguish the usage.  Makerspace model is used to MakerspaceView, while tool model is used to the tool view.  

Makerspace - A view of  description of makerverkstedet and href to tools
Tools -  A view of all available tools.

makerspace/tests.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+# Create your tests here.
+class ToolTestCase(TestCase):
+    def setUp(self):
+        self.first_tool = Makerspace()
+
+    def test_create_a_tool(self):
+        self.first_tool.title = 'test tool'
+        self.assertEqual(self.first_tool.title,'test tool')
```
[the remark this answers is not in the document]
qpb96: @siireen @HaraldWilhelmsen

makerspace/converters.py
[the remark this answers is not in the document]
qpb96: @siireen @HaraldWilhelmsen

web/templates/web/admin_panel.html
```diff
@@ -80,6 +80,31 @@ <h1>{% trans "Admin panel" %}</h1>
                     </a>
                 </div>
             {% endif %}
+
+            {% if user|has_any_event_permission or user.is_superuser %}
```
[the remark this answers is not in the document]
qpb96: @siireen could you check this out?

makerspace/models.py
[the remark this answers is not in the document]
zootos: I understand the benefits of distinguishing the different usages. However, the Makerspace model will only ever be used for one single object, which is what made me initially wonder about why the model exists. For what it is worth, I think that since the image is not used, the model is for all intents and purposes equivalent to the ContentBox model. This model is designed for purposes like this, and I would say that using it instead would be better.

makerspace/urls.py
```diff
@@ -0,0 +1,14 @@
...
+    path('', views.ViewMakerspaceView.as_view(), name='makerspace'),
+    path(r'<int:pk>/edit/', views.ViewAdminMakerspaceView.as_view(), name='makerspace/edit'),
+    path(r'tools/', views.ViewToolsView.as_view(), name='makerspace-tools'),
+    path(r'tool/<int:pk>/', views.ViewToolView.as_view(), name='makerspace-tool'),
```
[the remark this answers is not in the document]
qpb96: I think r'tools/' works as the same way as articles and news. 
A page to display all tools.
@siireen @HaraldWilhelmsen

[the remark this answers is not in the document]
ddabble: True - it doesn't affect the function in any way, it's just that the `r` is a bit unnecessary, and makes the paths look a little inconsistent 😛

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,40 @@
...
+{% load static %}
+{% load i18n %}
+
+{% block body %}
+    <h1 class="ui text-centered make_col_yellow header">
+        {% trans "Tools" %}
+    </h1>
+    <div class="ui text container">
+        <a class="ui yellow button" href="{% url 'makerspace-tools-create' %}">
+            <i class="ui add icon green"></i>
```
hermabe, line 11: The green color here is not used in these buttons elsewhere. Should be removed for consistency.

Maybe even remove the button entirely and use a plus icon by the "Tools" title similarly as for machines.

makerspace/templates/tools/tools.html
```diff
@@ -0,0 +1,47 @@
...
+
+{% block head %}
+    <link rel="stylesheet" type="text/css" href="{% static "tools/css/tools.css" %}">
+    <script src="{% static "web/js/image-link.js" %}"></script>
+{% endblock head %}
+
+
+{% block body %}
+    <h1 class="ui text-centered make_col_blue header">
+        {% trans "Available Tools at Makerverkstedet" %}
```
hermabe, line 14: This is missing a translation.

makerspace/migrations/0002_auto_20181004_1855.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 2.1.1 on 2018-10-04 16:55
```
hermabe, line 1: There are a lot of migrations here, could they be combined?

makerspace/templates/tools/admin_tool.html
```diff
@@ -0,0 +1,40 @@
...
+                {% for tool in tools_list %}
+                    <div class="item">
+                        <i class="ui icon"><img class="inline square_36" src="{{ tool.image.url }}"></i>
+                        <div class="content">
+                            <div class="header">
+                                <a class="inline-block make_col_blue header" href="{% url 'makerspace-tools-edit' tool.id %}">
+                                    {{ tool.title }}
+                                </a>
+                                {% if perms.makerspace.delete_tool %}
+                                    <a class="delete confirm inline-block make_col_blue explanation-popup"
```
hermabe, line 25: This popup is not shown.

makerspace/templates/tools/tool.html
```diff
@@ -0,0 +1,44 @@
...
+{% load static %}
+{% load i18n %}
+{% load thumbnail %}
+{% load permission_tags %}
+
+{% block title %}{{ tool.title }}{% endblock title %}
+
+{% block og_title %}{{ title.title }}{% endblock og_title %}
+{% block og_image %}{{ title.image.url }}{% endblock og_image %}
+{% block og_description %} En utstyrsside {% endblock og_description %}
```
hermabe, line 11: This should be translated.

makerspace/views.py
```diff
@@ -0,0 +1,58 @@
...
+    model = Tool
+    template_name = 'tools/tools.html'
+    context_object_name = 'tools_list'
+
+
+class AdminToolView(TemplateView):
+    template_name = 'tools/admin_tool.html'
+
+    def get_context_data(self, **kwargs):
+        if not has_any_tool_permission(self.request.user) and not self.request.user.is_superuser:
```
hermabe, line 25: The explicit check for superuser can be removed. A superuser has all permissions.

makerspace/views.py
```diff
@@ -0,0 +1,58 @@
...
+    fields = "__all__"
+    success_url = reverse_lazy('makerspace-tools-admin')
+
+
+class DeleteToolView(PermissionRequiredMixin, DeleteView):
+    model = Tool
+    success_url = reverse_lazy('makerspace-tools-admin')
+    permission_required = 'makerspace.delete_tool'
+
+    def get(self, request, *args, **argv):
```
hermabe, line 57: Why is GET implemented?

web/templates/web/admin_panel.html
```diff
@@ -92,6 +92,30 @@ <h1>{% trans "Admin panel" %}</h1>
                     </a>
                 </div>
             {% endif %}
+            {% if user|has_any_tools_permission or user.is_superuser %}
```
hermabe, line 95: Again, the superuser check.

makerspace/urls.py
```diff
@@ -0,0 +1,14 @@
+from django.urls import path
+from django.contrib.auth.decorators import login_required
+from makerspace.views import ToolsView, ToolView, EditToolView, CreateToolView, \
+    AdminToolView, DeleteToolView
+
+
+urlpatterns = [
```
hermabe, line 7: How does one reach these urls? I se no link in the header?

web/templates/web/admin_panel.html
```diff
@@ -92,6 +92,30 @@ <h1>{% trans "Admin panel" %}</h1>
...
+                        <div class="center aligned header">
+                            <i class="ui grey tools icon"></i>{% trans "Tools" %}
+                        </div>
+                        <div class="description">
+                            {% trans "Here you can add or edit tools at MAKE NTNU." %}
+                        </div>
+                    </a>
+                </div>
+            {% endif %}
+            {% if  user|has_any_makerspace_permission or user.is_superuser %}
```
hermabe, line 107: Superuser check.

web/templates/web/admin_panel.html
```diff
@@ -92,6 +92,30 @@ <h1>{% trans "Admin panel" %}</h1>
...
+                <div class="ui yellow card">
```
hermabe, line 108: Is this card needed? It just links to the publicly available makerspace info page.

locale/nb/LC_MESSAGES/django.po
```diff
@@ -2125,3 +2141,16 @@ msgstr "Velg verdi"
...
+msgid "Go to tools"
+msgstr "Gå til utstyrside"
+
+#: makerspace/templates/admin_tool.html:13
+msgid "Add new tool"
+msgstr "Legg til nytt verktøy"
+
+#: makerspace/templates/admin_tool.html:28
+msgid "Click to delete"
+msgstr "Trykk for å slette"
```
hermabe, line 2156: I think you forgot to run `compilemessages`.

makerspace/urls.py
Maisergodt: Editing or creating a tool can be accessed through the admin panel. Regarding the tool page, you cannot possibly reach it unless you have the correcet URL. Rather than linking to each tool which is what we are currently doing now, we could similarly refer to a overview page of all available tools instead:  `/makerspace/tools`.

makerspace/templates/tools/admin_tool.html
Maisergodt: It does show up for me
image

hermabe: I thought there was another popup, as only `delete confirm` is required for the deletion confirmation modal.

ddabble: Has this been resolved, @hermabe?

hermabe: `inline-block make_col_blue explanation-popup` is not needed for the delete modal. If there is another popup that should show it must be fixed. If not, remove the uneccessary classes.

ddabble: They've been removed in c8b990c3ae481c4c5bdc61fb59e106efbeb676c8 🙂 Resolving conversation.

makerspace/urls.py
ddabble: If I understand you correctly, @Maisergodt, should we add a link to `/makerspace/tools` on the `/makerspace` page? If that's the case, then it's fine by me 😊 What do you think, @hermabe?

hermabe: There needs to be a link the users/non-make-members can access the tools page. I didn't see a link. One could do it in the makerspace contentbox, but something more permanent may be better. I am not sure where it fits best. We could maybe add a dropdown in the header, but we haven't used that anywhere else so it breaks the pattern.

Maisergodt: Yes, I was thinking of adding a link in the makerspace contentbox page. Something like this such that regular users can access it as well. 
image 

I agree a permanent solution would be better. Although I am not certain where it's suitable.

sigtheidiot: I'm not against a link in the contentbox. If we were to have a permanent link i think it should be either above or below the contentbox.

sigtheidiot: The newest commits has added a dropdown menu with links to "makerverkstedet", "regler" and "verktøy". This is a solution to what was talked about here.

sigtheidiot (approved):

web/templates/web/header.html
```diff
@@ -72,9 +74,9 @@
         </a>
     </nav>
 
-    <nav id="side-nav" class="ui secondary huge inverted menu">
+    <div id="side-nav" class="ui secondary huge inverted menu">
```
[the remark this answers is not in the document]
ddabble: Functionally (for most users), it makes not difference, but it can make a difference for users with screen readers, as it adds a `role="navigation"` attribute (see [ARIA: Navigation Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#Description)). Additionally, there was another `nav` element nested inside this one - which caused there to be two of those `"navigation"` roles for the same set of navigation links, and the inner one contained the links the most "directly", so I thought this made the most sense 🙂 See [the docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) for more info.

PR #245, opened 2019-11-06
SLA print

Maisergodt:


web/settings.py
```diff
@@ -31,6 +31,15 @@
 EVENT_TICKET_EMAIL = "ticket@makentnu.no"
 EMAIL_SITE_URL = "https://makentnu.no"
 
+
+SENDGRID_API_KEY = os.getenv('SENDGRID_API_KEY')
+
+EMAIL_USE_TLS = True
+EMAIL_HOST = 'smtp.gmail.com'
+EMAIL_HOST_USER = 'test.makentnu@gmail.com'
+EMAIL_HOST_PASSWORD = 'makentnu'
```
ddabble: Never commit passwords or similar secrets, especially when this is a public repository. Please replace this line with something like
```
def get_email_password():
    email_secret_file = os.path.join(BASE_DIR, "email_secret.txt")
    try:
        with open(email_secret_file) as f:
            return f.read().strip()
    except IOError:
        if not DEBUG:
            raise FileNotFoundError(f"Please create the file \"{email_secret_file}\" with the password to {EMAIL_HOST_USER}")


EMAIL_HOST_PASSWORD = get_email_password()
```
and make sure to create the file `email_secret.txt` in the `BASE_DIR` of the server containing the password to test.makentnu@gmail.com.

kalkins: This should be done in the `local_settings.py` file instead, it's not tracked by git and used in addition to the main settings file. The server already has these variables set so it sends from `noreply@makentnu.no`

kalkins: So remove these lines, move them to `web/local_settings.py` locally, and change the password of the email account.

make_queue/views/reservation/reservation.py
```diff
@@ -14,6 +15,26 @@
...
+
+
+
+class SLA_request(TemplateView):
+    template_name = "make_queue/SLA.html"
+
+    def post(self, request):
+        self.contact_info = request.POST.get("email")
+        self.file = request.POST.get("stl_file")
+        self.content = request.POST.get("description")
```
kalkins: There's no need to use `self` here because the values are only used in this function. Use normal variables instead

requirements.txt
```diff
@@ -9,7 +9,7 @@ channels
 channels_redis
 django_ical
 xlsxwriter
-python_ldap==3.1.0
+python_ldap==3.2.0
```
kalkins: Why is this updated?

make_queue/views/reservation/reservation.py
```diff
@@ -14,6 +15,26 @@
...
+
+        msg = EmailMessage("SLA request", self.content + '\n' + self.contact_info, self.contact_info, ["test.makentnu@gmail.com"])
```
kalkins: The email address has to be changed from the testing one

make_queue/views/reservation/reservation.py
zootos: Also, wouldn't it be better to use a Django form here? Both for cleaner code and backend checks on e.g. if an actual email is given.

web/settings.py
zootos: The server may actually send from any `@makentnu.no` email address (even if the email address is not manually created). So this should probably be set to send from `sla-request@makentnu.no` or something similar.

make_queue/views/reservation/reservation.py
zootos: I think this might fail anyways, as it seems to try to send the message using the given email address. Which, if I'm not mistaken Google might not let us do.

make_queue/templates/make_queue/SLA.html
```diff
@@ -0,0 +1,28 @@
...
+             <input type="file" name="stl_file" accept=".stl" name="file" title="{% trans open %}">
+        </div>
+        <div>
+            <input class="ui blue submit button" type="submit" name="button" value="{% trans "Send" %}" >
+        </div>
+    </form>
+
+{% endblock body %}
+
+
```
zootos: Not important, but would be nice to remove all but one of these empty lines at the end of the file :)

zootos:
The email functionality of Django is not async, meaning that the way this code is implemented now will result in the user having to wait until the email is properly sent. While in most cases this is okay, it is better to send emails async. I actually implemented support for this a little way back. The documentation was not so good, and I did not implement support for attaching files. However, #246 should fix all of these issues. Personally, I think it would be better to implement the sending of the emails using this instead.

make_queue/templates/make_queue/reservation_machines.html
```diff
@@ -27,6 +27,10 @@
             <a href="{% url "find_free_slot" %}" class="ui primary button">
                 {% trans "Find free reservation slots" %}
             </a>
+            <a href="{% url "SLA_request" %}" class="ui primary button">
```
hermabe: Two things:
- The SLA printer is currently out of order, so it does not make sense to open up for requests. Maybe it should be possible to toggle the visibility of this button?
- As we add more functionality to the reservation system, we end up with more and more blue buttons here. It is ok for now with only three buttons, but I think we maybe should rethink the design here.

web/settings.py
kalkins: @zootos But this is the email used in the settings file, which shouldn't be changed from the default. If possible each different application should use a different appropriate address.

make_queue/views/reservation/reservation.py
kalkins: @zootos Would it work to use `noreply@makentnu.no` in the settings file, and override it with `sla-print@makentnu.no` here? As I understand it the email address used here was just for testing.

Anyway, the `reply-to` field should be set, as no one is receiving email on `sla-print@makentnu.no`.

web/settings.py
zootos: The `noreply@makentnu.no` defined for `EMAIL_HOST_USER` just defines the which GSuite user to use. I have whitelisted the IP of the server with the `noreply@makentnu.no` user in GSuite to be allowed to send emails from whatever `@makentnu.no` email trough the GSuite SMTP relay. So with the current configuration on the server, we can set the `from` field to be any `@makentnu.no` email.

make_queue/views/reservation/reservation.py
zootos: Yeah, see my new comment above :)

PR #248, opened 2019-11-16
Event signup

hermabe:
Requires users to log in to register for events. This enables us to disallow a user to register multiple times for the same event. If a user is not logged in, he will be redirected to the login page and then the registration form. When a user is registered for an event, the register button on the event page will become green and redirect to the "My tickets" page.


locale/nb/LC_MESSAGES/django.po
```diff
@@ -1626,13 +1631,19 @@ msgid ""
...
 msgstr ""
 "Det var ikke mulig å melde deg på arrangementet på dette tidspunktet. "
 "Vennligst prøv igjen ved en senere anledning."
 
-#: news/views.py:351
+#: news/views.py:348
+#, fuzzy
+#| msgid "No tickets have been registered for this event"
+msgid "You are already registered for this event."
+msgstr "Ingen billetter har blitt registrert for dette arrangement"
```
zootos: This fuzzy translation is not correct and needs to be changed

news/views.py
```diff
@@ -326,6 +326,12 @@ def event(self):
             return get_object_or_404(Event, pk=self.kwargs["event_pk"])
         return None
 
+    def get(self, request, *args, **kwargs):
```
zootos: Use [LoginRequiredMixin](https://docs.djangoproject.com/en/2.2/topics/auth/default/#the-loginrequired-mixin) instead. Then you don't have to check this and the same thing further down.

hermabe: That is better.

hermabe:
>The event page crashes when you are not logged in


How did i miss that? It should work now.

hermabe:
Also, when all old events that does not require user login have happened, we can remove the functionality for claiming tickets.

news/templates/news/event.html
```diff
@@ -75,7 +81,12 @@
             {% if article.standalone and article.timeplace_set.future %}
                 <div class="ui card">
                     {% if article.number_of_tickets %}
-                        {% if article.number_of_registered_tickets < article.number_of_tickets %}
+                        {% is_registered article user as registered %}
+                        {% if registered %}
+                            <a href="{% url "my-tickets" %}" class="ui green fluid button">
```
zootos: Wouldn't it be better to link to the page for that specific ticket?

news/views.py
```diff
@@ -335,13 +335,14 @@ def form_valid(self, form):
             form.add_error(None, _("Could not register you for the event, please try again later."))
             return self.form_invalid(form)
 
-        ticket = form.save()
-        if self.request.user.is_authenticated:
-            ticket.user = self.request.user
+        if EventTicket.objects.filter(user=self.request.user, active=True, timeplace=self.timeplace, event=self.event).exists():
+            form.add_error(None, _("You are already registered for this event."))
```
zootos: Instead of checking this and giving an error, can't we just redirect the user to their ticket if they for some reason try to access this view (both GET and POST)?

news/models.py
```diff
@@ -214,3 +214,13 @@ class Meta:
         permissions = (
             ("cancel_ticket", "Can cancel and reactivate all event tickets"),
         )
+
+    @property
+    def name(self):
+        # For backwards compatibility, name is no longer set
```
zootos: This comment should probably be above the `_name` field instead of here, with a docstring for what the function does, here instead. Same goes for the `email` property below with the `_email` field.

news/templates/news/event.html
hermabe: Yes, then it will be the same as the redirect after registering.

news/views.py
hermabe: Yes, there is no point in showing the form at all.

zootos:
> Also, when all old events that does not require user login have happened, we can remove the functionality for claiming tickets.

Could we have a PR on this ready? Just so that it is ready for deployment

hermabe:
Yes, I can make a separate PR for that.

PR #249, opened 2019-11-17
Remove claim ticket functionality

hermabe:
Removes functionality for users to claim tickets that are not registered to a user. This is not longer needed because we require users to log in before registering for events.

This PR must be merged after all the events with tickets without users have happened. We must thus merge #248 first, then wait. Currently the last event for this semester is November 25, and it is unlikely that there will be any more events until January.


zootos:
Now that all tickets will be registered to users, we could also require the user to be logged in to visit the ticket page.

PR #253, opened 2019-12-14
Replace Travis CI with Github Actions

kalkins:
I did this mostly to test out the new github actions, not because I really wanted to replace Travis.

The system seems like it's flexible and easy to use. The marketplace of free addons is great (like the codecov and cache actions), and it's nice to have everything collected here instead of having to go of to another site to check the build. Actions can also do a lot of other stuff, like reacting to comments.

The build itself seems to be slightly slower than with Travis, mostly because I had to install apt packages to make ldap work. @zootos how did this work on Travis without install more packages?

Note that notifications are controlled by your github settings, not the workflow file, and it defaults to sending emails for failed workflows. Go to 'settings->notifications->Github Actions' to change this.


zootos:
I don't remember, but I think we actually didn't have to do anything for Travis. Nonetheless, I agree that using GitHub is better than Travis in regards to gathering everything in a single system. Changed the base from `master` to `dev`, as we will have to get the changes into the `dev` branch at some point, and   the changes to the `dev` branch will trickle into the `master` branch at the next deploy

PR #255, opened 2019-12-17
Deployment

zootos:
Small deployment, so as to handle the non-standard changes of #240 by themselves. Also includes minor fixes from #254.


kalkins:
Looks good. @zootos update the server?

zootos:
> Looks good. @zootos update the server?

Yeah, I can fix that

PR #256, opened 2019-12-17
Documentation functionality

zootos:
Adds simple functionality for dev documentation on the subdomain `docs`. It does this by adding a simple wiki style app, which allows for creation of arbitrary pages having each page as a CKEditor editable page. Includes some basic functionality:

- Creation, editing and deletion of pages
- Page history (through each edit being a new page content DB object)
- Simple search

There is probably many thing that can be done to make this better. However, we greatly need a place to collect documentation. This is meant as a minimal starting solution, where changes can be made at a later date.


kalkins:
Great! I've been looking into this too, but I haven't come far in actually making it yet

kalkins:
Okay, I've looked through it, and I have a couple of comments:
- Why doesn't the search page link to the matching pages? Right now it seems impossible to navigate to a page without entering the url manually
- Having the name of the page in the url is nice for readability, but it risks dead links if titles are changed. Maybe we could use PKs in addition, and have a "get permanent link" on the top of the page?
- We need a way to navigate to the documentation from the main page. I'm guessing you either forgot, or didn't add it yet because it's a work in progress?

kalkins:
I enabled a plugin in CKEditor to add code snippets wirh syntax highlighting, which is essential for programming documentation

kalkins:
BTW, right now the front page can be deleted, basically bricking the whole documentation...

kalkins:
I also have some suggestions for additional features. Most of these could take a long time, and should be done as separate PRs later, but I list them here anyway:

- Creating categories for pages, and a box with a list of collapsing categories with pages on the left side
- Mentor would probably like to have public documentation for 3D-printers and such. Could we make a subset of the wiki public? Maybe have a setting on the category, and showing that on the main page?
- I think we need more features in CKEditor to make it easy to make pages that look good. There's a lot of functionality built in, we only have to enable and configure it

kalkins:
I just have to say, I really like the way you did the history of each page, it works perfectly

hermabe:
>Mentor would probably like to have public documentation for 3D-printers and such. Could we make a subset of the wiki public? Maybe have a setting on the category, and showing that on the main page?

I think the Tool pages will cover what we need to have public. However, it would be nice to have a place for internal guides. There are some maintenance and fixing that only MAKE members should do.

Categories for sorting 3Dprinter stuff from dev stuff would be nice.

zootos:
> * Why doesn't the search page link to the matching pages? Right now it seems impossible to navigate to a page without entering the url manually

It is just an oversight. Meant to add it, but forgot to do so.

> * Having the name of the page in the url is nice for readability, but it risks dead links if titles are changed. Maybe we could use PKs in addition, and have a "get permanent link" on the top of the page?

Right now there is no ability to rename titles. I see that it might be useful to change the title of a page. Would it be an idea to allow renaming of titles, but add URLs both for titles and PKs, with the PK one redirecting to the one with the title (as you suggest). Further creating a simple CKEditor plugin, which lets you search for the tile of a page, and then it adds the PK link to the page?

> * We need a way to navigate to the documentation from the main page. I'm guessing you either forgot, or didn't add it yet because it's a work in progress?

My intent was to link to other pages of documentation in the content of the main page, with the search allowing for quick navigation.

> BTW, right now the front page can be deleted, basically bricking the whole documentation...

I now disallowed deletion of the main page.

> * Creating categories for pages, and a box with a list of collapsing categories with pages on the left side

Categories would be good, but I they are not necessary for the first version.

> * Mentor would probably like to have public documentation for 3D-printers and such. Could we make a subset of the wiki public? Maybe have a setting on the category, and showing that on the main page?

Based on Hermans answer it seems like it would not be needed, at least for now, to add public pages. I would, however, suggest that we could actually split the Wiki in two, one under `docs.` for dev documentation (which is not needed by the rest of MAKE) and one under `wiki.` for everything else. We could use the same templates, views and models. The only difference would be giving each page an extra field indicating if it is for the `wiki.` or `docs.` subdomains. With the titles being unique per subdomain.

> * I think we need more features in CKEditor to make it easy to make pages that look good. There's a lot of functionality built in, we only have to enable and configure it

Absolutely! It will probably be easier to find out which plugins are needed after using the functionality for a while. We can also create our own plugins, if we want.

kalkins:
> Right now there is no ability to rename titles. I see that it might be useful to change the title of a page. Would it be an idea to allow renaming of titles, but add URLs both for titles and PKs, with the PK one redirecting to the one with the title (as you suggest). Further creating a simple CKEditor plugin, which lets you search for the title of a page, and then it adds the PK link to the page?

Yes, a redirect would probably be best. A CKEditor plugin would be great to easily reference pages without having to open it up and copy the link. Anyway, this is not important right now, and can be added later.

> > * We need a way to navigate to the documentation from the main page. I'm guessing you either forgot, or didn't add it yet because it's a work in progress?
> 
> My intent was to link to other pages of documentation in the content of the main page, with the search allowing for quick navigation.

Sorry, I was a bit vague. I meant that I don't see a way to navigate from the public page to the docs subdomain, or any way to get out of the subdomain. While we're on this, I don't see why docs is under its own subdomain, instead of under `internal.`. As I see it the only reason to put something in its own subdomain is because its very separated from other functionality, and needs its own navigation bar. I don't think this applies to docs.
 
> Based on Hermans answer it seems like it would not be needed, at least for now, to add public pages. I would, however, suggest that we could actually split the Wiki in two, one under `docs.` for dev documentation (which is not needed by the rest of MAKE) and one under `wiki.` for everything else. We could use the same templates, views and models. The only difference would be giving each page an extra field indicating if it is for the `wiki.` or `docs.` subdomains. With the titles being unique per subdomain.

I would prefer using categories to separate this, instead of having them on different parts of the site. If people don't need the dev docs they can keep that category collapsed. However, if they do need it (like if they want to learn about django or the camera system) it's easily available, they don't have to dig through the site.

I'll approve this now, so we can get it to prod as quick as possible. I'll start writing docs for dev right away, and then we can add more features in new PRs.

kalkins:
@zootos Could you explain that last commit? What do you mean by "content is not changed across articles"?

zootos:
> Sorry, I was a bit vague. I meant that I don't see a way to navigate from the public page to the docs subdomain, or any way to get out of the subdomain. While we're on this, I don't see why docs is under its own subdomain, instead of under internal.. As I see it the only reason to put something in its own subdomain is because its very separated from other functionality, and needs its own navigation bar. I don't think this applies to docs.

I added it under a separate subdomain, as I initially thought of it as only dev documentation (and thus separate from everything else). However, I think that keeping it 100% separate (without a link), would be a good choice while it is in this rough state. Then, when we want to open it fully to all of MAKE, we can move it to `internal` and add links.

>> Based on Hermans answer it seems like it would not be needed, at least for now, to add public pages. I would, however, suggest that we could actually split the Wiki in two, one under docs. for dev documentation (which is not needed by the rest of MAKE) and one under wiki. for everything else. We could use the same templates, views and models. The only difference would be giving each page an extra field indicating if it is for the wiki. or docs. subdomains. With the titles being unique per subdomain.

> I would prefer using categories to separate this, instead of having them on different parts of the site. If people don't need the dev docs they can keep that category collapsed. However, if they do need it (like if they want to learn about django or the camera system) it's easily available, they don't have to dig through the site.

Yeah, this might be better.

kalkins:
Okay, sounds good

zootos:
> @zootos Could you explain that last commit? What do you mean by "content is not changed across articles"?

It should have said "pages", my bad. The point here, is to make sure that we don't get pages that use a content object that does not point to them. That is, the commit simply remove the ability for crafting requests that set the content of a page as  the content of another page. If we allowed this (and somebody made bad requests), then we would easily end up with non-valid database states (e.g. history would not work).

zootos:
The commit also removes all but the one choice that will be used from the hidden form for changing the version of the page (as to reduce the amount of HTML that will be created and sent when there are a lot of content objects)

zootos:
If you are happy with the last commit, I will merge and then create a PR to `master`

kalkins:
Ahh, ok. I can't see anyone crafting requests like that on purpose, but it's good to guard against it. Just in theory, we could still use the admin page like this to make a page that uses the content for a previous version of another page?

I'm happy, merge when ready. You can go straight to deployment too if you want.

zootos:
Yeah, we can do that. Just a guard against any theoretical possible "attacks".

PR #260, opened 2020-01-27
Fix/reservation error messages

sigtheidiot:
Add more reservation error messages and translations


make_queue/views/reservation/reservation.py
```diff
@@ -29,10 +30,23 @@ def get_error_message(self, form, reservation):
...
+            return _("Reservations can only be made {:} days ahead of time".format(
+                reservation.reservation_future_limit_days));
         if self.request.user.has_perm("make_queue.can_create_event_reservation") and form.cleaned_data["event"]:
-            return "Tidspunktet eller eventen, er ikke lengre tilgjengelig"
-        return "Tidspunktet er ikke lengre tilgjengelig"
+            """Translation: Tidspunktet eller eventen, er ikke lengre tilgjengelig"""
+            return _("The time slot or event, is no longer available")
+        if not reservation.quota_can_make_reservation():
+            """Translation: Reservasjonen går over kvoten"""
+            return _("The reservation exceeds the quota")
```
zootos: A bit pedantic, but wouldn't it be better with `"The reservation exceeds your quota"`?

make_queue/views/reservation/reservation.py
```diff
@@ -29,10 +30,23 @@ def get_error_message(self, form, reservation):
         """
         if not reservation.is_within_allowed_period_for_reservation() and not (
                 reservation.special or reservation.event):
-            return f"Reservasjoner kan bare lages {reservation.reservation_future_limit_days} dager frem i tid"
+            """Translation: Reservasjoner kan bare lages {:} dager frem i tid"""
+            return _("Reservations can only be made {:} days ahead of time".format(
+                reservation.reservation_future_limit_days));
```
ddabble: The closing parenthesis of `_()` should be after the end of the translation string, not after the call to `format()`. Also, the colon in `{:}` and the semicolon at the end of the statement are not needed 😋

I would suggest one of these two alternatives:
* `_("Reservations can only be made {} days ahead of time").format(reservation.reservation_future_limit_days)`
* `_("Reservations can only be made {num_days} days ahead of time").format(num_days=reservation.reservation_future_limit_days)`

The last one follows a nice convention of naming the placeholder variable(s), which helps the translators know what the type of the value is, though it's probably kind of obvious in this case.

make_queue/models/models.py
```diff
@@ -208,8 +208,21 @@ def validate(self):
...
 
         # Check if the user can make the given reservation/edit
+        return self.quota_can_make_reservation()
+
+    # Check if the start time is before current time
+    def reservation_starts_before_now(self):
+        return self.start_time < timezone.now()
+
+    # Check if start time is after end time
+    def check_start_time_after_end_time(self):
```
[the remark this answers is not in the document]
sigtheidiot: The reason it's outside of the verify method is so i could call it in the reservation file and use it to give the connected error message.

make_queue/views/reservation/reservation.py
sigtheidiot: Yeah, that sounds better

make_queue/views/reservation/reservation.py
kalkins: Even better, we can [pluralize](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#pluralization) like this:
```python
from django.utils.translation import ngettext

num_days = reservation.reservation_future_limit_days
return ngettext(
    'Reservations can only be made {num_days} day ahead of time',
    'Reservations can only be made {num_days} days ahead of time',
    num_days
).format(num_days=num_days)
```

locale/nb/LC_MESSAGES/django.po
```diff
@@ -191,6 +192,12 @@ msgstr "Et eller flere språk mangler innhold"
 msgid "Save"
 msgstr "Lagre"
 
+#: docs/forms.py:42
+#, fuzzy
+#| msgid "No content exists for this page"
```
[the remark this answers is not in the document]
kalkins: Just to explain what happened here: When a string is slightly changed gettext tries to identify what the old version was, and replace that. In this case it thinks that the old version was "No content exists for this page", and that the new version is "The content does not belong to the given page". If this is correct, you can remove the line with `fuzzy` and the old version. However, gettext can make mistakes, so check that it's correct

make_queue/views/reservation/reservation.py
```diff
@@ -29,10 +30,23 @@ def get_error_message(self, form, reservation):
         """
         if not reservation.is_within_allowed_period_for_reservation() and not (
                 reservation.special or reservation.event):
-            return f"Reservasjoner kan bare lages {reservation.reservation_future_limit_days} dager frem i tid"
+            """Translation: Reservasjoner kan bare lages {:} dager frem i tid"""
```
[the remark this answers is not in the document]
kalkins: Note that if you want to keep them, or even when just using them temporarily, it's best to start it with the word `Translators`. Django recognizes that keyword in comments one line above a translation, and copies it to the `.po` file for easy reference. More info: https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#comments-for-translators

make_queue/models/models.py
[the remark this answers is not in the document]
kalkins: I think it's fine to have it outside, even though it's a bit simple and short

make_queue/models/models.py
```diff
@@ -208,8 +208,21 @@ def validate(self):
             return False
 
         # Check if the user can make the given reservation/edit
+        return self.quota_can_make_reservation()
+
+    # Check if the start time is before current time
+    def reservation_starts_before_now(self):
+        return self.start_time < timezone.now()
+
+    # Check if start time is after end time
```
kalkins: This comment should be a docstring instead. Generally, when a comment describes a function or class, it should be a docstring. You can make it like this:
```python
def check_start_time_after_end_time(self):
    """Check if start time is after end time"""
```

More info on docstrings: https://www.python.org/dev/peps/pep-0257/

kalkins:
Also, remember to run `python manage.py compilemessages` after changing either the english or norwegian part of the translations. You didn't do that after your last change.

make_queue/views/reservation/reservation.py
```diff
@@ -182,7 +202,8 @@ def dispatch(self, request, *args, **kwargs):
         :param request: The HTTP request
         """
         # User must be able to change the given reservation
-        if not kwargs["reservation"].can_change(request.user) and not kwargs["reservation"].can_change_end_time(
+        if not kwargs["reservation"].can_change(
+                request.user) and not kwargs["reservation"].can_change_end_time(
                 request.user):
```
kalkins: The way this line is separated makes it a bit hard to read. I would suggest changing it to this instead:
```python
if not kwargs["reservation"].can_change(request.user)
   and not kwargs["reservation"].can_change_end_time(request.user):
```
or
```python
if not (kwargs["reservation"].can_change(request.user)
        and kwargs["reservation"].can_change_end_time(request.user)):
```

make_queue/views/reservation/reservation.py
```diff
@@ -226,7 +247,8 @@ def post(self, request, *args, **kwargs):
             return self.get(request, *args, **kwargs)
 
         reservation = reservations.first()
-        if not reservation.can_change_end_time(request.user) or reservation.start_time >= timezone.now():
+        if not reservation.can_change_end_time(
+                request.user) or reservation.start_time >= timezone.now():
```
kalkins: I think this is easier to read:
```python
if not reservation.can_change_end_time(request.user)
   or reservation.start_time >= timezone.now():
```

PR #262, opened 2020-02-01
Sign up for old events

zootos:
Fixes a bug where it was possible to sign up for old standalone events. Also tries to make it easier to see which events have already happened.


news/views.py
```diff
@@ -87,6 +87,8 @@ def get_context_data(self, **kwargs):
         context.update({
             'article': event,
             'timeplaces': event.timeplace_set.all() if event.standalone else event.timeplace_set.future(),
+            'is_old': not event.timeplace_set.future().exists(),
+            'last_occurrence': event.get_past_occurrences().exists() and event.get_past_occurrences().first() or None
```
ddabble: This way of writing a conditional assignment is pretty hard to read; please replace it with
```python
event.get_past_occurrences().first() if event.get_past_occurrences().exists() else None
```

zootos: This is just me not thinking. Actually, `first()` returns `None` if there are no objects in the queryset, so no conditional is even needed.

news/models.py
```diff
@@ -135,10 +135,10 @@ def can_register(self, user):
             return False
         if self.private and not user.has_perm("news.can_view_private"):
             return False
-        if self.standalone:
-            return self.number_of_tickets > self.number_of_registered_tickets()
         if not self.get_future_occurrences():
             return False
+        if self.standalone:
```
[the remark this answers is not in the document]
zootos: This is actually the fix to the problem this PR addresses, where standalone events would just check the number of tickets and not if there are any future occurrences. I actually moved the `if not self.get_future_occurrences():` statement, but how you change the order is the same to git. I didn't add a comment specifying the importance of the order, but I made comments to indicate the function of each if-statement. So, I think it should be clear now that they should be in this order. Actually, the order of the 3 first if-statements does not matter, as they all just return `False`.

zootos:
I fixed the translations. They actually belong to some old changes, that I think forgot to update the translations (they where fuzzy translations and just new lines added by `makemessages`).

PR #263, opened 2020-02-01
Reduction of JPEG sizes

zootos:
Lately I have been looking at improving the performance of the website. Through configuring Nginx caching policies and compression of text in transfer, I have been able to reduce the data transferred at first load of the homepage from 2.5 MB to 1.6 MB. Of these 1.6 MB, 1.1 MB are JPEG images. Some of these are quite small manually configured images hardcoded in HTML. These only account for about 0.1 MB of data. The remaining 1 MB of data are images from articles. Opening the articles page, you have to load an even larger amount of data.

This PR overrides the saving of articles and events, saving all JPEG images with a quality level of 90 (100 is the best). For images that are already below or at this quality level, this does not change anything. For images above this quality level, a large reduction in file size will be achieved. This will result in a slightly lower quality of these images, however going from quality 100 to quality 90 results in only minuscule changes in quality, that at worst are barely visible. For almost all images, the difference is not at all noticeable.

Taking the front page as an example, this would reduce the size of the page by between 0.7 MB and 0.8 MB. Effectively halving the amount of data transferred. The articles page would have an even larger reduction of transferred data.

### DIFF
--- a/news/models.py
+++ b/news/models.py
@@ -82,15 +82,19 @@ class NewsBase(models.Model):
         """
         # Only check the image if there is actually an image
         if self.image:
-            image = Image.open(self.image)
-            if image.format == "JPEG":
-                output = BytesIO()
-                image.save(output, format="JPEG", quality=90)
-                output.seek(0)
+            # PIL will throw an IO error if it cannot open the image, or does not support the given format
+            try:
+                image = Image.open(self.image)
+                if image.format == "JPEG":
+                    output = BytesIO()
+                    image.save(output, format="JPEG", quality=90)
+                    output.seek(0)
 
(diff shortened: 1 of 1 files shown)


news/models.py
```diff
@@ -70,6 +74,30 @@ class NewsBase(models.Model):
...
+                    image.save(output, format="JPEG", quality=90)
+                    output.seek(0)
+
+                    self.image = InMemoryUploadedFile(output, "ImageField", self.image.name, "image/jpeg",
+                                                  sys.getsizeof(output), None)
+                image.close()
+            except IOError:
+                pass
+
+        super(NewsBase, self).save(kwargs)
```
ddabble: The `**` before `kwargs` are missing 😋

PR #264, opened 2020-02-02
Maintenance reservations

hermabe:
Implements maintenance reservations. Maintenance reservations will show up as a separate type in the calendar and the status of the machine is set.

As there is no longer a need for the `status` field on the `Machine` model, it is replaced by a `out_of_order` boolean field. All the other status choices is found by checking the current reservation.

As the status of the machine is no longer stored in the database, an enum is used for the status choices. The text representation can now easily be found by `machine.status.value`.


zootos:
I haven't checked everything, but I think it might be smart to do some more changes to the Reservation model. When the Reservation model was originally created, I think I only had events as a special type. At that point, it made sense to just have an event field in the model. With the expansion to MAKE NTNU reservations, I was a little unsure, but I went for just another field. However, for each field we create, the queries, and checks become more and more obfuscated and complex. So I would think it might be better to do the following instead of just adding another field for maintenance reservations;

1. Create a field for reservation type in the Reservation model with four options. (STANDARD, EVENT, MAKE NTNU, MAINTENANCE)
2. Remove the boolean fields for MAINTENANCE and MAKE NTNU
3. Remove the text field called `special_text` and keep its content in the `comment` field. (There are a couple reservations where both of these fields are used, but they basically say the same)
4. Update the form to have a drop down instead of checkboxes (this is only showed based on permissions)
5. Simplify all the other code based on this.

I realise that this is quite a big job, and I can do it myself, but I think the reduced complexity is worth the time.

hermabe:
I agree with the simplifications. We will likely be adding more functionality to the reservation system, eg. checkin or recurring reservations. Can be good to clean up a bit before adding any more.

I just did the simplest thing to add maintenance reservations, but at some point some refactoring is due.

hermabe:
There are some simplifications to the `Machine` model here which I can move to a separate branch (without the maintenance reservations).

zootos:
> There are some simplifications to the `Machine` model here which I can move to a separate branch (without the maintenance reservations).

If we are going to go do the suggested changes, I think this would be a good idea. Doing things in steps will simplify the review process, and probably reduce the amount of resulting errors.

hermabe:
The changes to the `Machine` model is now on `simplify/machine`.

ddabble:
> I realise that this is quite a big job, and I can do it myself, but I think the reduced complexity is worth the time. So I am mainly interested in hearing what you guys think about it @hermabe @kalkins @ddabble @sigtheidiot?

Those 5 changes do absolutely sound reasonable, so if you decide to start working on this, it would be great! 😊

sigtheidiot:
> I realise that this is quite a big job, and I can do it myself, but I think the reduced complexity is worth the time. So I am mainly interested in hearing what you guys think about it @hermabe @kalkins @ddabble @sigtheidiot?

I think it sounds like a good plan!

zootos:
Just to keep this PR up to date, #280 implements a new reservation calendar. I intended to start implementing the changes discussed above by changing the add/edit reservation page to use more of the builtin Django form functionality (due to later changes that must be made to the form). However, the form was page was tightly connected to the reservation calendar, which made it hard to change the add/edit reservation page alone. Thus, it will probably be a while before the changes discussed above are implemented (other things should/need to be fixed/changed before one makes these changes).

hermabe:
Closing this. Will reimplement after the refactoring.

ddabble:
@hermabe can the `feature/maintenance-reservations` branch be deleted, or does it contain work that's planned to be implemented in the future?

PR #272, opened 2020-02-15
Simplify articles and events

zootos:
Changes articles and events to use datetime fields instead of separate fields for date and time. This is done for timeplace start and end time as well publication time for both articles and timeplaces. The main reason for this change is to simplify the code. It might seem like I'm adding more code than I am removing. However, a lot of the changes are due to migrations, adding a widget for a datetime input using the Fomantic UI style and translations. There are many more changes that can be done to simplify the code even more, but these should be made in other PRs, as to be able to see what changes are actually made and easier detect errors.

This code also implements a check in the form for timeplaces, so that the timeplace can no longer end before it starts.

Also added a missing info message on old events on mobile layout.


news/models.py
```diff
@@ -18,27 +16,19 @@
...
-        return self.published().filter(
-            Q(end_date=timezone.localtime().date(), end_time__gt=timezone.localtime().time()) |
-            Q(end_date__gt=timezone.localtime().date()))
+        return self.published().filter(start_time__gte=timezone.localtime())
 
     def past(self):
-        return self.published().filter(
-            Q(end_date=timezone.localtime().date(), end_time__lt=timezone.localtime().time()) |
-            Q(end_date__lt=timezone.localtime().date()))
+        return self.published().filter(start_time__lt=timezone.localtime())
```
ddabble: Was it on purpose that `past()` and `future()` now filter relative to start instead of end time..? 😮

news/templatetags/news_tags.py
```diff
@@ -12,15 +11,9 @@ def color(val):
...
+    return timeplace_set.filter(end_time__lt=timezone.now())
 
 
 @register.filter(name="future")
 def future(timeplace_set):
-    return timeplace_set.filter(Q(end_date=timezone.now().date(), end_time__gt=timezone.now().time()) |
-                                Q(end_date__gt=timezone.now().date()) |
-                                Q(end_date=None, end_time__gt=timezone.now().time(), start_date=timezone.now().date()) |
-                                Q(end_date=None, start_date__gt=timezone.now().date()))
+    return timeplace_set.filter(end_time__gt=timezone.now())
```
ddabble: One of `past()` and `future()` should probably be `lte` or `gte`, like you changed with `TimePlaceManager.past()` and `TimePlaceManager.future()` 🙂

news/migrations/0017_use_datetime_instead_of_date_and_time.py
```diff
@@ -0,0 +1,124 @@
...
+        article.save()
+
+
+def merge_timeplace_date_and_time(apps, schema_editor):
+    Timeplace = apps.get_model("news", "Timeplace")
+    for timeplace in Timeplace.objects.all():
+        timeplace.publication_time = datetime.combine(timeplace.pub_date, timeplace.pub_time)
+        timeplace.start_time = datetime.combine(timeplace.start_date, timeplace.temp_start_time)
+        timeplace.end_time = datetime.combine(
+            timeplace.end_date if timeplace.end_date is not None else timeplace.start_date, timeplace.temp_end_time)
```
ddabble: Curious why `start_date` and `temp_end_time` are combined when there is no `end_date`.. Because it's the closest we can get to a datetime that seems about right..? 🤔

ddabble:
Added a commit fixing a missing info message from #262; hope that kind of fits with this PR 😬

news/migrations/0017_use_datetime_instead_of_date_and_time.py
zootos: From what I remember when Amund implemented this 3 years ago was that `end_date` would be `None` only when the start and end date was the same. Thus I have chosen to combine `start_date` with `temp_end_time` in those cases.

news/models.py
zootos: No, it is just a mistake from on my part. End time is indeed better.

PR #277, opened 2020-02-22
Misc changes to articles and events

zootos:
I have kept a list of smaller changes and improvements I would have liked to do to the admin and article pages. This PR contains many of these smaller changes to the pages:
- Removes weird space between the "Attributes" label and checkboxes in the event and article forms 
- Orders events by last occurrence in the admin panel
- Simplifications of the cards used in the admin pages for articles and events. This is done to reduce the space each article/event uses
- Misc missing and/or wrong translations
- Sort previous occurrences by newest first in the admin panel for an event
- Fix some missing changes from date to datetime field in templates for timeplaces
- Fix an error in the image compression, which hindered PNG images from being uploaded
- Fix the overlap between the calendar icon and long dates on event occurrences
- Show number of registered tickets to admins (based on permissions) in the normal event page
- Changes to how an occurrence looks on mobile


news/templates/news/admin_events.html
```diff
@@ -15,50 +15,49 @@ <h1 class="ui text-centered header">
...
                         <div class="meta">
-                            {% if event.get_future_occurrences %}
-                                {{ event.get_future_occurrences.count }} {% trans "Future occurrences" %}
-                            {% endif %}
-                        </div>
-                        <div class="description">{{ event.clickbait }}</div>
-                    </div>
-                    {% if event.featured or event.hidden or event.private %}
-                        <div class="extra content">
+                            {% with event.get_future_occurrences.count as num_future_occurrences %}
```
hermabe: This will only count published `TimePlace`s. I was confused for a bit because it showed no future occurences even though the event had one. The event was hidden so `get_future_occurrences()` returned nothing.

Maybe count all occurrences?

zootos: I hadn't really considered that, as I just reused the code from earlier, but that is better. I have changed it now.

ddabble:
Added some commits!

Btw, I assume that this branch was renamed from [`fix/misc-articles-and-events`](https://github.com/MAKENTNU/web/commit/b316e4a86a9d205c4f97fccd3b2ece5789f3f95c) (last updated 05.10.2019)..? In that case, can we delete the old branch? 🙂

zootos:
> In that case, can we delete the old branch? 🙂
Yeah, I started on these changes this fall, but stopped for some reason. It was easier to start anew than fix merge conflicts. I will delete the other branch.

PR #278, opened 2020-02-29
Fix the add and edit member pages

zootos:


ddabble:
We should probably write more tests, at least one for each endpoint 😬

hermabe:
> We should probably write more tests, at least one for each endpoint 😬

Yes, I clearly did not test this well enough when implementing it.

PR #280, opened 2020-03-08
New Async Reservation Calendar

zootos:
To simplify implementation of changes discussed in #264 I have chosen to reimplement the reservation calendar mostly using Javascript with asynchronous requests and objects. This simplifies the flow and implementation, removing the dependencies of using templates and/or loading new HTML each time a different machine/week should be shown. Especially, I tried to make the calendar as self-contained as possible, such as to simplify and streamline interaction between it and the rest of the code.

There is a lot of code and a lot of changes. I have done my best to test everything, but I might have missed something, or possibly files that can now be deleted. 

I tried to restrict the changes to the files and pages that used the reservation calendar to only the things that concern the calendar itself, as to limit the scope of the PR. Other changes will probably be made in the future as a part of the changes discussed in #264.

The new calendar, in addition to being asynchronous and easier to use in implementation, contains several fixes/improvements:
- Changing the selection time in increments of 5 minutes
- Working properly for users that can ignore the rules
- Allowing for selection over multiple days (within the same week)
- Setting the selection to the start/end of the given day when the headers and footer of the table are hovered.
- Automatically updating the red bar used for time indication (once per minute)
- Fixing an issue where the selection could be longer than the user was allowed
- Fix an issue where the selection popup was not working correctly in Safari (discovered a year ago!)


zootos:
I can add that the file `reservation_rules.js` is mostly just a copy of the file `reservation_rules.html`. The only difference is a change from `toPrecision` to `toFixed` to fix a bug.

hermabe:
Looks good! One small thing:
I think the popup looks better if the padding is set to zero:
image
As opposed to a popup with a card inside:
image

zootos:
> Looks good! One small thing:
> I think the popup looks better if the padding is set to zero:
> image
> As opposed to a popup with a card inside:
> image

Absolutely, this looks much better. I made a change that removes the padding for both this and the button on the make/edit reservation pages. I kept the card class on the popup to make it sufficiently large, as otherwise it felt quite cramped. 

I guess this was the only thing you wanted fixing, and that the PR is okay to merge now?

PR #285, opened 2020-03-12
Global message about the COVID-19 closing

zootos:
A quick and dirty global message about the closing of Makerverkstedet. This should be made into a standard system, where messages can be created without change to the code. The current message is just meant as a stopgap solution.


zootos:
> Good. We should probably get a system so that these types of messages can be set from the admin panel sometime in the future.

That is the plan, we just need to push this message out first :)

PR #287, opened 2020-03-14
Announcements

zootos:
Adds a simple announcement system to the website. Allows for creation of three different types of announcements (INFO: blue, WARNING: yellow, ERROR: red), that can be set to show on either just the front page or on all pages. Support multiple announcements being shown at the same time, as well as having support for timed and indefinite announcements.


zootos:
I had a bit of trouble deciding on the names to use for the different types of announcements. I am open for changes to the current names.

announcements/templates/announcements/create_announcement.html
```diff
@@ -0,0 +1,65 @@
...
+        <form class="ui form" method="POST" action="">
+            {% csrf_token %}
+            <div class="ui two fields">
+                <div class="ui field">
+                    <label>{{ form.classification.label }}</label>
+                    {{ form.classification }}
+                </div>
+                <div class="ui field">
+                    <label>{{ form.link.label }}
+                        <span data-content="{% trans "An optional link to an information page" %}">
```
hermabe: Maybe use the `help_text` Django field attribute for these?

locale/nb/LC_MESSAGES/django.po
```diff
@@ -1976,42 +2058,37 @@ msgstr "Se gammelt bilde"
...
-
 #: web/widgets.py:16 web/widgets.py:26
 msgid "Choose value"
 msgstr "Velg verdi"
 
 #: web/widgets.py:51
 msgid "Search places"
 msgstr "Søk etter steder"
+
+#~ msgid "Error"
```
hermabe: These are not used

zootos:
> Looks good. Just a couple small comments. Also, some translations were missing, but was fixed by running `compilemessages`.

I was changing the translations around in the end, and guess I missed a compilemessages after the last change or two.

> "Info" and "Warning" names are good. "Error" is kind of misleading as there are no errors. Maybe something like "Critical", I'm not sure.

Critical is better. I will change to it for now, and then maybe somebody has a better suggestion

announcements/templates/announcements/create_announcement.html
zootos: Absolutely! I have never seen that field before, TIL.

PR #290, opened 2020-03-15
Fix HTTP 500 when changing the retired/quit status of an already changed member

zootos:
Fixes an issue, where if two people (or one person with two windows) sat the retired/quit status of a member after each other, then the second person would get a HTTP 500 error. We now instead gracefully fail, using Django messages to tell the user that their action was not performed due to the status already being changed.


zootos:
> I think you forgot to add the new translation strings in `django.po` (and `django.mo`) 😅

Seems so... (:

PR #291, opened 2020-03-15
Fix event ticket email template

zootos:
HTML email rendering is not very standardized, and while the current template used for emails works quite well for several clients, it does not work perfectly for all email clients. In some cases padding and/or elements may disappear or get rendered randomly. This PR tries to fix this problem by using HTML tables, which are much more generally supported by the clients than other HTML elements. Further, more hard-coding of CSS properties is done to make sure that the less random rendering choices are made by the email clients.

Also adds the event image to the email.


zootos:
The branch is running on the dev server, so checking the email in your email clients can easily be done by signing up to the "open" event on there.

PR #292, opened 2020-04-04
Feature privacy policy context box

sigtheidiot:
Added a content box for our privacy policy and a link to the content box from the footer.


web/static/web/css/footer.css
```diff
@@ -91,3 +91,7 @@
         grid-column-start: auto;
     }
 }
+
+#footer .content {
+    text-align: right;
+}
```
ddabble, line 97: This should be moved to a `@media screen and (min-width: 993px)` block, so that it doesn't apply on mobile layout 🙂 A suggestion is to place that block before the `max-width: 992px` block, to be consistent with e.g. `header.css`.

If you want, you could also add a comment above the `max-width: 992px` block saying something like `/* Switch to mobile layout */`, which would make the CSS file more readable 😊

Lastly, remember to include a newline at the end of a file 😬 It doesn't make any difference in practice, but it's a convention 😅

ddabble: On second thought - after having looked at the other contents of this file, it's probably better to simply move the `text-align: right;` property to the `#footer` block at line 5 😁

ddabble: On second thought - after having looked at the rest of the contents of this file, it's probably better to instead add this block to the `@media screen and (max-width: 992px)` block:
```
#footer .content {
    text-align: center;
}
```
And also move the existing `#footer .content` block to just below e.g. the `#footer .social.list i` block (both to let blocks further down the file overwrite the properties of the blocks above, and to group blocks affecting neighboring elements).

ddabble:
Just changed the order of two CSS blocks in that last commit 🙂 

Also, I noticed that you added an extra newline at the end of the `footer.css` file, so that there were two newlines 😅 In case that was because you didn't see any newline at the end of the file on GitHub, I thought I might inform you that GitHub doesn't show the last newline of a file - only a  icon if it's missing 😛 It is visible, however, if you open the file in a text editor, like GitHub's integrated one.

sigtheidiot:
> Just changed the order of two CSS blocks in that last commit 🙂
> 
> Also, I noticed that you added an extra newline at the end of the `footer.css` file, so that there were two newlines 😅 In case that was because you didn't see any newline at the end of the file on GitHub, I thought I might inform you that GitHub doesn't show the last newline of a file - only a  icon if it's missing 😛 It is visible, however, if you open the file in a text editor, like GitHub's integrated one.

Thank you, I did not know that, but i do now.

PR #293, opened 2020-04-08
Catch broader Dataporten exception

ddabble:


kalkins:
Wouldn't it be better to catch all exceptions, just in case? Also, there is `logging.exception` which automatically logs details about the exception, including a stacktrace: https://docs.python.org/3/library/logging.html#logging.Logger.exception

kalkins:
@zootos  @ddabble ?

ddabble:
Oh, sorry! I agree that we could just as well catch all exceptions, and that it's better to use `logging.exception`. I was planning on implementing some additional logging, but I guess we can finish this PR first 😅

ddabble:
Added the `exc_info` kwarg, to avoid PyCharm's warning of "Too broad exception clause" when the caught `Exception` is not used.

PR #297, opened 2020-04-26
Fix login redirect when changing language

ddabble:
**Summary:** Replaces the custom-written `decorated_includes()` with `decorator_include()` of the library `django-decorator-include`.

This fixes a bug that caused anonymous users to get redirected to login when trying to change the site's language. This could be reproduced by:
1. *If not logged in since server restart:* Log in and then log out.
2. Press the button for changing language in the header.

This bug was caused by `decorated_includes()` decorating `django.conf.urls.i18n.set_language()` with `permission_required("internal.is_internal")`. (This was done from `internal/urls.py`, but still affected `web/urls.py`.)


zootos:
I like the idea of using a library for this instead of some homemade solution. It seems like the chosen library has not been updated in 13 months, and is in general no longer maintained. I have yet to test the code, but am a bit sceptical about introducing a library that seems to no longer be maintained and that is not meant for the major version of Django that we are running.

kalkins:
I agree with @zootos. The implementation in the library is very short, maybe we could copy it and maintain it ourself if necessary?

ddabble:
@zootos I agree, and shared the same skepticism when I noticed that it was a bit old.. I would still prefer it over the current solution, though, as it works properly in all areas as far as I can tell, and looks like it's well tested - albeit not for the current major Django version, like you pointed out 😅 

@kalkins Regarding maintaining the repo: I was planning on opening a PR which would simply upgrade the test suite to test the newest version of Django (and Python, for that matter), and see if any code actually needs updating. It seems to me like perhaps the 3.0 update of Django didn't cause backwards incompatibility with the parts of the framework that this repo uses 🤔

zootos:
I agree that it is better than the current solution. Nonetheless, I think we should chose one of the options that fix the aforementioned problems before merging the PR. 

Due to the license used for the library, I think the updating of the repo with a PR will probably be the simplest solution, although I am open to both.

ddabble:
@zootos @kalkins [django-decorator-include](https://pypi.org/project/django-decorator-include/) was updated in May, and now officially supports Django 3.0. So it seems like it's occasionally maintained 😁

PR #298, opened 2020-04-26
Fix GitHub workflow

ddabble:
* Added a trigger on `pull_request`, in case we get PRs from forks
* Now testing the two newest Python versions
* Fixed uploading report to Codecov
* Upgraded a couple action versions

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,7 @@
+[run]
+# Current, top-level directory
+source=.
+
 [report]
 exclude_lines =
     # Ignore not abstract methods, as these cannot be tested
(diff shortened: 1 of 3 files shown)


sigtheidiot:
> Looks good to me. I haven't used GitHub workflow before, so I think it is better if @kalkins takes a look and approves it.

Same here, I also have no experience with this so can't give a good review

ddabble:
I agree, but I can't seem to find anything that only triggers on PRs from forks..

However, if only the amount of CPU time we get for free is considered, I don't think it matters that much - fortunately, as I'm pretty sure that there are usually a lot more pushes than there are opened PRs 😅

PR #304, opened 2020-08-09
Docker container for development

Maisergodt:
# Docker container proposal

## Background
I had some trouble installing the dependencies (python_ldap), and I assume new members will share the same experience as well. 

## Solution
Docker container with all dependencies installed, and docker is the only thing you need to install. Docker and containers are more and more widespread, and we should probably also start using it.

I have created a docker file that creates a docker container with all dependencies and a docker-compose file to run it.

## How-to
- [Install docker](https://docs.docker.com/get-docker/)
- Run the command `docker-compose up` in the root directory of the repository.
- After it has downloaded and built, it should be available on the [makentnu.localhost:8000](http://makentnu.localhost:8000) if that is what is specified on the left side of the `:` in the `docker-compose.yml` file:
```yml
ports:
    - "8000:8000"
```
- Code changes should automatically update, as normal.

## Notes
This is for development use. We could also set up a deployment container, which copies the files, instead of sharing the folder. It would be cool to set up a CI/CD pipeline with this as well.


ddabble:
Just for future reference: `python-ldap` on Windows can be installed by downloading a wheel from [this page](https://www.lfd.uci.edu/~gohlke/pythonlibs/#python-ldap) (mentioned in the [documentation](https://www.python-ldap.org/en/python-ldap-3.3.0/installing.html#windows)) 🙂

kalkins:
Oops, I started to comment on this weeks ago, but apparently it didn't get sent.

This is a nice change, especially for those using Windows. Later we can extend the usage of Docker to include build checks and deployment. One of the advantages with Docker is that it's easy to set up a database, like Postgresql, instead of sqlite like we're using now. This gives better performance and more features.

I suggest that we merge this soon, and add more features later.

kalkins (approved):

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3
+ENV PYTHONUNBUFFERED 1
+RUN apt-get update && apt-get install -y python-dev libldap2-dev libsasl2-dev libssl-dev
```
mahoyen, line 3: `gettext` is needed to be able to generate translations.

```suggestion
RUN apt-get update && apt-get install -y python-dev libldap2-dev libsasl2-dev libssl-dev gettext libgettextpo-dev
```

mahoyen (approved):
I like it!

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3
+ENV PYTHONUNBUFFERED 1
+RUN apt-get update --fix-missing && apt-get install -y python-dev libldap2-dev libsasl2-dev libssl-dev gettext libgettextpo-dev
+RUN mkdir /web
+WORKDIR /web
+COPY requirements.txt /web/
+RUN pip install -r requirements.txt
```
ddabble, line 7: I think it's good practice to update `pip` before installing anything - which is what we do in the CI file ([`python.yml`](https://github.com/MAKENTNU/web/blob/b886772aaa580b8f7c2b397b60086c0be2ec131d/.github/workflows/python.yml#L32)) as well 😊 Though I'm unsure if `python` will use the intended Python version, or if we have to write `python3` or `python310` 🤔
```suggestion
COPY requirements.txt /web/
RUN python -m pip install --upgrade pip
RUN pip install -r requirements.txt
```

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+services:
+  web:
+    build: .
+    command: >
+        bash -c "python manage.py migrate
+        && python manage.py runserver 0.0.0.0:8000"
+    volumes:
+      - ./:/web
+    ports:
+      - "8000:8000"
```
ddabble, line 12: Missing newline :p

ddabble (approved):
As we have talked about earlier, I was initially hesitant with merging this PR, as I thought we should add more configuration to the Dockerfile so that it could potentially mirror the setup we have on the website server, and keeping this PR open would kind of serve as a reminder for that. Now, even though that would still be nice, no one has even started working on that, and it feels pretty silly keeping the PR open for this long, so let's just merge it 😅 Having these files locally, might also make it easier for people to start experimenting with Docker, which might even lead to the more complete Dockerfile mentioned earlier 🤠
So sorry that it took this long for me to make up my mind about this ❤️

Lastly, since it's been almost 2 years, is there anything that should be updated in these files, like versions or if there are any new, preferred ways of doing things..? 🤔

Dockerfile
Maisergodt: Tried to ouput the version numbers for different python versions, and can confirm that it does indeed use the intended version :)

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3
```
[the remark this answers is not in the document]
Maisergodt: It will use the latest `3.x.x` version, but probably better to restraint it to `3.10.x` to avoid any compatible issues whenever a newer version is released 

https://docs.docker.com/engine/reference/builder/#official-releases

Maisergodt:
> Lastly, since it's been almost 2 years, is there anything that should be updated in these files, like versions or if there are any new, preferred ways of doing things..? 🤔

Besides the suggestions you recommended, I don't think there is anything else to change 😊

ddabble (approved):
Alright, thanks a bunch! 😄

PR #305, opened 2020-08-09
Fix dead MazeMap link

ddabble:


ddabble:
> I think you forgot to update the po file, since the second translation is now on a different line?

Are you referring to updating the line number in the comments in the po file? If so, it's not necessary for the translations to work, but I can commit it for documentation's sake, if you want 🙂

sigtheidiot:
> > I think you forgot to update the po file, since the second translation is now on a different line?
> 
> Are you referring to updating the line number in the comments in the po file? If so, it's not necessary for the translations to work, but I can commit it for documentation's sake, if you want 🙂

Okay, good to know :) then you don't have to do it

PR #309, opened 2020-09-09
Move events before articles

sigtheidiot:
This PR moves events so they appear before articles on the front page. It also reduces the height of the MAKENTNU banner.


ddabble (approved):
Very nice! 🔥

PR #310, opened 2020-09-11
Fix reservation "change" button

sigtheidiot:
The "change" button on the "my reservations" page would not respond if the reservation had begun. This PR lets users change the reservation end time and comment after the reservation has begun.


make_queue/views/reservation/reservation.py
```diff
@@ -217,6 +217,10 @@ def form_valid(self, form, **kwargs):
         if reservation.machine != form.cleaned_data["machine"]:
             return redirect("my_reservations")
 
+        # If the reservation has begun, the user is not allowed to change the start time
+        if reservation.start_time < timezone.now() and reservation.start_time != form.cleaned_data["start_time"]:
+            return redirect("my_reservations")
```
ddabble, line 222: This redirect (and the one above it on line 218) provides no feedback to the user about what happened, so I think we should change it so that the user stays on the change reservation page and an error message is shown.

If this were to be implemented, I think we should do the following:
* Move the validation logic in this view to the form's `clean()` method (where all validation usually occurs).
* Either change `ReservationForm` to inherit from `ModelForm` instead of just `Form`, as it ideally should, though this could potentially be a fairly large refactor (haven't looked at the code a lot). If not, 
I think we could pass the `Reservation` instance to the form (on line 128) as a kwarg (e.g. named `instance`), saving it in the form's `__init__()` method, check whether it's `None` or not in the form's `clean()` method, and then run the validation logic in the step above (raising `ValidationError` if it's invalid).
* If the form is invalid, ensure that an error message is shown after returning from the `handle_post()` method (line 122).

Any thoughts? 😊

ddabble (changes requested):
The code looks good, though I have commented a few improvements to (mainly the old) code, which I think would be beneficial 😄

make_queue/views/reservation/reservation.py
hermabe: I think maybe @zootos started som refactoring of the reservation code. I think the canges to the `ReservationForm` was planned. I do not know how far he got though.

make_queue/views/reservation/reservation.py
```diff
@@ -202,7 +202,7 @@ def dispatch(self, request, *args, **kwargs):
         """
         # User must be able to change the given reservation
         if not (kwargs["reservation"].can_change(request.user)
-                and kwargs["reservation"].can_change_end_time(request.user)):
+                or kwargs["reservation"].can_change_end_time(request.user)):
             return redirect("my_reservations")
         return super().dispatch(request, *args, **kwargs)
```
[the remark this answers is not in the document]
sigtheidiot: Yes, that is definitely easier to read

make_queue/views/reservation/reservation.py
sigtheidiot: Yeah, feedback should be implemented, but right now the front-end of the website "stops" the user from changing the start time after the reservation has started, and this was just to have a backend stop for it too. 

But the reservation code should be refactored in the future.

ddabble: Sure, this can be done in the future 👌😊

ddabble (approved):

PR #312, opened 2020-09-23
Stop people from booking machines listed as "Out of Order"

sigtheidiot:
This PR includes checks that will stop people from booking printers that are marked as out of order


Maisergodt:
Should probably also prevent the users from booking when the machines are under maintenance. Currently it is a possbile consequence.

make_queue/templates/make_queue/reservation_actions.html
```diff
@@ -3,7 +3,8 @@
 
 <h3>{% trans "Actions" %}</h3>
 <div class="ui relaxed divided selection action list">
-    <a class="item {% if not can_use or not can_make_more_reservations %}disable{% endif %} enable_popup"
+    {% machine_out_of_order machine as machine_out_of_order %}
+    <a class="item {% if not can_use or machine_out_of_order or not can_make_more_reservations %}disable{% endif %} enable_popup"
```
ddabble, line 7: I think something like this would be a bit more readable:
```diff
-    {% machine_out_of_order machine as machine_out_of_order %}
-    <a class="item {% if not can_use or machine_out_of_order or not can_make_more_reservations %}disable{% endif %} enable_popup"
+    <a class="item enable_popup
+              {% if not can_use or machine.get_status == Machine.OUT_OF_ORDER or not can_make_more_reservations %}
+                  disable
+              {% endif %}"
+       href="{% url 'make_reservation' machine=machine %}"
             {% if not user.is_authenticated %}
        data-content="{% trans "You must be logged in to make new reservations" %}"
             {% elif not can_use %}
        data-content="{% get_machine_cannot_use_text machine %}"
             {% elif not can_make_more_reservations %}
        data-content="{% trans "You have reached the maximum number of future reservations" %}"
             {% endif %}
-       href="{% url 'make_reservation' machine=machine %}">
+    >
```
This involves removing the `machine_out_of_order` template tag you created - and yes, I'm aware that there exists a lot of template tags which are in the same style as this one, but, frankly, they're all slightly superfluous, when one can simply use the methods that are part of the model that people are more likely to be familiar with. (E.g. why does the `get_machine_cannot_use_text` template tag exist, when one can simply write `machine.machine_type.cannot_use_text`? It kind of baffles me 😅) And if there is a certain piece of code related to a model object that is commonly written, one could just create a method/property for it on the model itself 😊

Also, my suggested code requires adding this line to [the template context](https://github.com/MAKENTNU/web/blob/b503ec2e53de0ce7027f025075196305cf45a7db/make_queue/views/reservation/calendar.py#L21-L30):
```Python
"Machine": Machine,
```

make_queue/templatetags/reservation_extra.py
```diff
@@ -90,6 +90,11 @@ def can_use_machine(machine, user):
     return machine.can_user_use(user)
 
 
+@register.simple_tag()
+def machine_out_of_order(machine):
+    return machine.get_status() == "O"
```
ddabble, line 95: See the above comment.

ddabble (changes requested):
Looks good! A couple comments, however 🙂

make_queue/templates/make_queue/reservation_actions.html
sigtheidiot: This looks like a much better solution, and yes, i dont totally understand why we use all of the template tags cx

ddabble: Hahah, yeah, and there are so many of them 😅😵

But, yeah, if we want to add more machine statuses that should prevent the user from creating a new reservation (like if the machine is under maintenance, as @Maisergodt mentioned above), it might be a fitting opportunity for creating a model method, like `Machine.is_reservable()`, which could check `get_status()` 😄

sigtheidiot: Ooo good idea. I've already pushed the changes you asked for, but i can do this too and append it ;)

mahoyen:
Great that you added a test for it! However, I noticed you added it after you fixed the bug. Have you seen if the tests fails when you comment out the essential changes? This way we know that the test tests the correct thing.

Normally you can avoid this extra loop by writing the test first, seeing it fail, and THEN fix the bug, and see the test pass :)

mahoyen:
> Should probably also prevent the users from booking when the machines are under maintenance. Currently it is a possbile consequence.

But isn't maintainance for a known period of time? So people should still be able to reserve timeslots for the week ahead, even though the printer is currently under maintainance?

hermabe:
> But isn't maintainance for a known period of time? So people should still be able to reserve timeslots for the week ahead, even though the printer is currently under maintainance?

Yes. There were plans to change the reservation system so that "maintenance" was a reservation type. I Implemented it in #264 but we decided to do some refactoring before adding new functionality to the reservation system.

ddabble:
> Great that you added a test for it! However, I noticed you added it after you fixed the bug. Have you seen if the tests fails when you comment out the essential changes? This way we know that the test tests the correct thing.
> 
> Normally you can avoid this extra loop by writing the test first, seeing it fail, and THEN fix the bug, and see the test pass :)

For all we know, that might have been exactly what she did 😛 Regardless, while I agree with your sentiment (i.e. regression testing), this is not very important with regards to the focus of this PR, which is simply to fix a bug that was observed. The test suite was lacking before this PR, and the test added does contribute towards fixing that, which I think is good enough for the scope of this PR 🙂

> But isn't maintainance for a known period of time? So people should still be able to reserve timeslots for the week ahead, even though the printer is currently under maintainance?

That's true, but we still need to fix (I presume..?) users being able to create a reservation which is within the maintenance period of a machine 🙂

sigtheidiot:
> > Should probably also prevent the users from booking when the machines are under maintenance. Currently it is a possbile consequence.
> 
> But isn't maintainance for a known period of time? So people should still be able to reserve timeslots for the week ahead, even though the printer is currently under maintainance?

There is a difference between booking a machine for maintainance, and setting the machine status to maintainance. This fix will only stop people from booking if the status is set to maintainance.

ddabble (approved):
Very nice! 😄

PR #314, opened 2020-10-13
Machine type model

ddabble:
It's now possible to add machine types through the Django admin site. Machine types (and machines) are now sorted by their priority; machines are sorted by name if priority is not specified.


sigtheidiot (approved):
Looks good to me!

Maisergodt (approved):

PR #322, opened 2020-10-26
Add equipment priority

ddabble:
This works in the same way as `Machine`'s `priority` field.


sigtheidiot (approved):
Looks good. I see that there's no stopping people from putting negative numbers in the priority field, don't know if you've given this any thought or not. But no need to change it, since it works well with negative numbers, and I have no argument for not using negative numbers.

ddabble:
@sigtheidiot Yeah, only positive numbers was intended, as you can imagine 😅 Negative numbers work just as well, though, so if users choose to input that, then that's fine, I guess..? 😛

PR #324, opened 2020-10-30
Internal mainpage

sigtheidiot:
This PR adds an internal mainpage. Right now the main page is empty, but we can add a picture or something more later


internal/templates/internal/home.html
```diff
@@ -1,5 +1,16 @@
...
+{% endblock %}
 
 {% block body %}
+   <div class="main">
+
+    <h1>{% trans "MAKE NTNU Internal" %}</h1>
+
+    </div>
 
 {% endblock %}
```
ddabble, line 16: It would be great if you could clean up this formatting 🙂 I can recommend regularly using a tool - like PyCharm's "Reformat Code" feature (Ctrl+Alt+L by default) - to keep a consistent code format.

```suggestion
    <div class="main">

        <h1>{% trans "MAKE NTNU Internal" %}</h1>

    </div>
{% endblock %}
```

internal/urls.py
```diff
@@ -19,7 +19,7 @@
     path("members/<int:pk>/retire", permission_required("internal.can_edit_group_membership")(MemberRetireView.as_view()), name="member-retire"),
     path("members/<int:pk>/retire/undo", permission_required("internal.can_edit_group_membership")(MemberUndoRetireView.as_view()), name="member-undo-retire"),
     path("members/access/<int:pk>/change", ToggleSystemAccessView.as_view(), name="toggle-system-access"),
-    # TODO: Change to the HomeView when there is actually some content to show there
+    path("", Home.as_view(), name="home"),
     path("", MembersListView.as_view(), name="home"),
```
ddabble, line 23: I didn't notice before now, but there are now two views with the same name, `home` 😅 Following the instructions of the removed comment, the following should be done:
```suggestion
    path("", HomeView.as_view(), name="home"),
```

`Home` should also be renamed to `HomeView`, as I've done above, to make it clearer that it's a view and not, say, a model 😛 (I know the view was created a long time ago, but it would be nice if the name could be improved as part of this PR 🙂)

sigtheidiot: Definitely , thank you. I must have missed the duplication in the merge process.

ddabble (approved):
Nice! 😁

PR #326, opened 2020-11-06
Copy Participants Emails

mahoyen:
There has long been a feature request from mentor to send email to all the participants of the event. This PR is meant offer a quick fix until the proper implementation is done.

This PR enables event holders to copy a list of participant emails, so that messages can easily be sent to participants.

The thought here is for the event holder to copy the list and paste it in their email client, and then they can write their email and send it there.

Preview:
image

### Translations
I don't know how to do this. Especially the 'Successfully copied to clipboard!' part.

### Tests
I want to write tests for
- Given active participant, then they are included in the list.
- Given participant inactive, then don't include them in the list.
- When the copy button is pressed, then the list is actually copied to clipboard.

Does django support writing these types of tests?


sigtheidiot:
I can fix the translations, and i think django can support the tests, but im not sure how 😅

locale/nb/LC_MESSAGES/django.po
```diff
@@ -1754,28 +1754,36 @@ msgstr "Ingen fremtidige forekomster finnes for dette arrangementet"
...
 msgid "No tickets have been registered for this event"
 msgstr "Ingen billetter har blitt registrert for dette arrangement"
 
+#: news/templates/news/admin_event_tickets.html:55
+msgid "Emails"
+msgstr "E-poster"
+
+#: news/templates/news/admin_event_tickets.html:57
+msgid "Click to copy email of active participants to clipboard"
+msgstr "Klikk her for å kopiere eposter til aktive deltagere til utklipstavlen"
```
Maisergodt, line 1785: Utklippstavle skrives med to p-er

locale/nb/LC_MESSAGES/djangojs.po
```diff
@@ -13,92 +13,51 @@ msgstr ""
...
 msgid "Comment"
 msgstr "Kommentar"
 
+#: make_queue/static/make_queue/js/make_reservation.js:278
+msgid "Choose time"
+msgstr "Velg tid"
+
+#: news/static/news/js/email_copy.js:21
+msgid "Successfully copied to clipboard!"
+msgstr "Kopiert til utklipstavlen"
```
Maisergodt, line 54: Two p's here as well.

news/tests.py
```diff
@@ -214,6 +220,15 @@ def toggle(pk, attr):
         self.assertEquals(toggle(self.article.pk, 'hidden'), {'color': 'grey' if hidden else 'yellow'})
         self.assertEquals(toggle(self.article.pk, 'hidden'), {'color': 'yellow' if hidden else 'grey'})
 
+    def test_ticketholder(self):
+        self.add_permission("change_event")
+        self.ticketholders[0].active = False
+        expected_string = ','.join([ticketholder.email for ticketholder in self.ticketholders if ticketholder.active])
+        response = self.client.get(reverse('timeplace-tickets', args=[self.timeplace.pk]))
```
mahoyen, line 227: For some reason this returns 404, and I don't know why

sigtheidiot:
image
After your commits from one hour ago, the emails no longer show up, and the "copy to clipboard" button no longers copies anything

mahoyen:
Yes, I tried using context data instead of creating it in the template, because that feels like the right way to do it and it is easier to test. Though I am having trouble getting it to work. It seems that the "_email" attribute is empty, which is why the emails do not show up. I don't know how to fix it yet.

mahoyen:
Now the functionality is restored. Still needs tests though. Thanks for fixing the translation!

news/tests.py
mahoyen: I found the problem. Since I had not defined `number_of_tickets` in the timeplace, it just returned 404.

news/static/news/css/admin-event-tickets.css
```diff
@@ -0,0 +1,4 @@
+.input-monospace {
+    font-family: 'Courier New', Courier, monospace;
+    width: 100%;
+}
```
sigtheidiot, line 4: Please add a new line at the end of the file

news/static/news/js/email_copy.js
```diff
@@ -0,0 +1,31 @@
...
+                on: 'manual',
+                exclusive: true
+            })
+            .popup('show')
+        ;
+        // Hide popup after 5 seconds
+        delayPopup(this);
+    });
+
+});
```
sigtheidiot, line 31: Here you're also missing a new line

news/tests.py
```diff
@@ -303,3 +327,19 @@ def test_not_published_article(self):
...
+        email = f"{username}@example.com"
+        user = User.objects.create_user(username=username,email=email)
+        if isinstance(event, Event):
+            ticketholder = EventTicket.objects.create(user=user, event=event, active=randint(0,1))
+        elif isinstance(event, TimePlace):
+            ticketholder = EventTicket.objects.create(user=user, timeplace=event, active=randint(0,1))
+        else:
+            raise NotImplementedError()
+        ticketholders.append(ticketholder)
+    return ticketholders
```
sigtheidiot, line 345: And here

sigtheidiot (approved):
Things look good! I've added a few comments that would be nice if you fixed before merging.

news/static/news/css/admin-event-tickets.css
mahoyen: Fixed! Why should I add it, though?

ddabble: It's just a convention; see the answers to [this question](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file) 🙂

mahoyen:
I am not authorized to merge, not sure if I am supposed to be. Please also note that I moved the email list to before the list of tickets, in case the list of tickets is long.

ddabble:
Sorry for being late to the party, but I have some things that I would prefer to be fixed before merging 😬

news/tests.py
```diff
@@ -100,7 +102,10 @@ def setUp(self):
         self.event = Event.objects.create(
             title='FUTURE',
             image=SimpleUploadedFile(name='img.jpg', content=simple_jpg, content_type='image/jpeg'),
+            number_of_tickets=40,
         )
+        self.timeplace = TimePlace.objects.create(event=self.event, start_time=timezone.localtime() + timedelta(minutes=5),
+                                      end_time=timezone.localtime() + timedelta(minutes=10), number_of_tickets=29)
```
ddabble, line 108: I think wrapping the code like this is more readable: 😊 
```suggestion
        self.timeplace = TimePlace.objects.create(
            event=self.event, number_of_tickets=29,
            start_time=timezone.localtime() + timedelta(minutes=5),
            end_time=timezone.localtime() + timedelta(minutes=10),
        )
```

ddabble (changes requested):
Please address the issues I've commented on below 🙂 The code looks very good otherwise!

news/templates/news/admin_event_tickets.html
```diff
@@ -1,6 +1,12 @@
 {% extends "web/base.html" %}
+{% load static %}
 {% load i18n %}
 
+{% block head %}
+    <script type="text/javascript" src="{% static "news/js/email_copy.js" %}"></script>
+    <link rel="stylesheet" type="text/css" href="{% static "news/css/admin-event-tickets.css" %}">
```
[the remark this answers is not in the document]
mahoyen: Here I followed what was already in the css folder. [admin-event.css](https://github.com/MAKENTNU/web/blob/feature/copy-participants-emails/news/static/news/css/admin-event.css) was there from before. Should I change both of them?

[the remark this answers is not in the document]
ddabble: Ah, I see. Renaming the other file isn't directly relevant to this PR, so you can leave it be 🙂 Besides, I'm planning on renaming it in an upcoming PR 😛

news/tests.py
```diff
@@ -303,3 +327,20 @@ def test_not_published_article(self):
...
+        email = f"{username}@example.com"
+        user = User.objects.create_user(username=username,email=email)
+        if isinstance(event, Event):
+            ticketholder = EventTicket.objects.create(user=user, event=event, active=randint(0,1))
+        elif isinstance(event, TimePlace):
+            ticketholder = EventTicket.objects.create(user=user, timeplace=event, active=randint(0,1))
+        else:
+            raise NotImplementedError()
+        ticketholders.append(ticketholder)
+    return ticketholders
```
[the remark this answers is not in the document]
mahoyen: I have done something similar. I put the `user_and_ticket_state_tuples ` as an input parameter, so that it is more general. Nice trick with the kwargs dictionary, I had not thought of that! Thanks :)

locale/nb/LC_MESSAGES/djangojs.po
```diff
@@ -13,92 +13,51 @@ msgstr ""
...
-msgstr "Torsdag"
-
-#: make_queue/static/js/calendar.js:67
-msgid "Friday"
-msgstr "Fredag"
-
-#: make_queue/static/js/calendar.js:67
-msgid "Saturday"
-msgstr "Lørdag"
-
```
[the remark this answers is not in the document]
sigtheidiot: Oh yeah, that makes a lot of sense. I will fix it as soon as I can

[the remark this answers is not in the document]
ddabble: I pushed a fix, so no worries 😅

news/tests.py
```diff
@@ -214,6 +221,47 @@ def toggle(pk, attr):
         self.assertEquals(toggle(self.article.pk, 'hidden'), {'color': 'grey' if hidden else 'yellow'})
         self.assertEquals(toggle(self.article.pk, 'hidden'), {'color': 'yellow' if hidden else 'grey'})
 
+    def test_event_ticket_emails_only_returns_active_tickeholders(self):
+        username_and_ticket_state_tuples = [
+            ("user1", False),
+            ("user1", True),
```
ddabble, line 227: The tests look much better now! I'm unsure, however, as to why you chose to write `"user1"` twice in this test (and the one below)..? This will create two tickets for the same user (which should be disallowed), and now when I read the value of `expected_string`, I can't really know which ticket the `user1@example.com` email comes from... 🤔 I'd suggest removing the first tuple in this list 🙂

mahoyen: It should probably be two separate tests. The scenario here is that a person cancels their original ticket, and then signs up again.

`get_or_create` was used because the user does not exist, except for `user1` the second time. So it was used to catch both cases. If I swap it out with `get` I get the error `User matching query does not exist`.

ddabble: Oh, sorry, I meant `User.objects.create_user()` 😅

news/tests.py
```diff
@@ -214,6 +221,80 @@ def toggle(pk, attr):
...
+            event=self.timeplace,
+            username_and_ticket_state_tuples=username_and_ticket_state_tuples
+        )
+        expected_ticket_emails = "user2@example.com,user4@example.com"
+        self.add_permission("change_event")
+        
+        response = self.client.get(reverse('timeplace-tickets', args=[self.timeplace.pk]))
+        
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(expected_ticket_emails, response.context["ticket_emails"])
```
mahoyen, line 260: There is a lot of code duplication here. Is there a way of parametrising the test? The only difference between these two tests are swapping out `event` for `timeplace`. I know this is possible with pytest, but it [does not look compatible](https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses) with `unittest.TestCase`

news/tests.py
mahoyen: I have added in the special case as a separate test, so I am keeping the `get_or_create`. As you can see from @sigtheidiot's [screenshot](https://github.com/MAKENTNU/web/pull/326#issuecomment-723480904). It is entirely possible to have two tickets for the same user.

ddabble: Cancelling a ticket is done by setting an existing ticket's `active` field to `False`, but the test code you've written simply creates two `EventTicket`s - one active and one inactive. It would absolutely be valuable to write a test for the `CancelTicketView`, though 😊 I don't see any value in writing a test that creates a ticket, sets the ticket's `active` field to `False`, and then checks the response, if that's what you're getting at, as that would pretty much be the same as just writing `False` in one of the tuples of `username_and_ticket_state_tuples` 🤔

> As you can see from @sigtheidiot's screenshot. It is entirely possible to have two tickets for the same user.

Yes, I know that it's possible, though it *shouldn't* be 😛 This isn't relevant to this PR, however.

news/tests.py
ddabble: I agree, and I guess the easiest thing to do would be to simply extract the common code to a separate method with an `event` and a `url_name` parameter (for the call to `reverse()`), and a name starting with e.g. `_test_...` (as methods not starting with `test` are not run), and calling this method from the two methods you've highlighted.

Also, while we're talking about it, it would be nice if you could move the `create_tickets_for()` function to just below the methods that are using it, indenting it (so that it's part of the `ViewTestCase` class), and annotating it with `@staticmethod` 🙂 It's easier to read the code if the methods calling each other are next to one another.

news/tests.py
mahoyen: So you should not be able to reregister an event? Or are you saying that if the person reregisters, they should keep their original ticket, where `active` is set to `True` again?

news/tests.py
mahoyen: I created a method called `assert_context_ticket_emails` that does most of the work. This method is called by the original tests. Even though it did not start with test, it looks like it runs, because it was called from the original test functions.

news/tests.py
ddabble: I just noticed that I hadn't read the existing code as thoroughly as I thought, so the tests you wrote makes sense now 🙂

> Or are you saying that if the person reregisters, they should keep their original ticket, where active is set to True again?

Yes, that sounds the most sane to me, as I don't really see any value in keeping a "log" of the times a user cancels and re-registers for an event 🤔

news/tests.py
ddabble: Nice! 😄 

> Even though it did not start with test, it looks like it runs, because it was called from the original test functions.

When I wrote "run", I meant that the method is not called directly by the test runner; however, any test method calling that function will naturally make it run, so yes, you're right 🙂

ddabble:
@mahoyen If you're not planning on making any additional changes, I'll merge this PR 🙂

mahoyen:
[This](https://github.com/MAKENTNU/web/pull/326#discussion_r519823542) still needs to be resolved. The debated tests are still in

ddabble:
As I mentioned in [my comment](https://github.com/MAKENTNU/web/pull/326#discussion_r519939665), the tests you wrote make sense to me now 🙂 I was waiting for you to reply to it, but I presume that you're content as well, so I'll mark it as resolved 👍

ddabble (approved):
Great! 🔥

PR #327, opened 2020-11-09
Some simplifiactions to the machine model

hermabe:
These are some simplifications I made when implementing maintenance reservations for almost a year ago. We decided then that we should do some refactoring before adding maintenance reservations. See #264 . We can maybe merge these now and add maintenance reservations later.

Changes:
 - Remove the status field and add an `out_of_order` field instead.
 - "Maintenance" is removed. It is not a machine property, but should be a reservation type.
 - Use an Enum for the different machine statuses


ddabble:
The model changes are very nice; It's been a slight annoyance of mine for some time that most of the status types do not need to be saved in a field, but can rather be derived from existing reservations, as the changes of this PR solidify! 🎉

I would, however, like to see that maintenance reservations are implemented before merging this PR, so we don't lose functionality (i.e. informing users) on the live website... What do you think about this, @hermabe @sigtheidiot @Maisergodt?

hermabe:
> I would, however, like to see that maintenance reservations are implemented before merging this PR

I had an implementation of this in #264 you can look at. There is likely changes needed to be done from what I had a year ago though.

Also, how often is the maintenance functionality used now? Currently, there is no difference between "out of order" and "maintenance".

sigtheidiot:
> Also, how often is the maintenance functionality used now? Currently, there is no difference between "out of order" and "maintenance".

There is no real difference functionality vice, but I agree that we should wait to merge this PR until #264 (or an updated version) is opened. I think it will make it easier for the users to understand what they're supposed to do if a machine is under matinence.

ddabble:
_Summarizing the most recent discussion above:_ Implementing #505 should be done before merging this PR.

PR #328, opened 2020-11-10
Decouple stream name from printer name

mahoyen:
This PR decouples the stream name/websocket url from the printer name.

## Background
Today the stream name/websocket url is the same as the printer name. This means that if we have to change switch the camera setup, the Raspberry Pi has to be reconfigured.

## Solution
By having a different stream name, the stream can be configured to belong to a different printer on the website and not on the Raspberry Pi.

I have added a new parameter `stream_name` to the machine model, and added an attribute to the img field called `data-stream`, which the [stream.js](https://github.com/MAKENTNU/web/blob/fix/decouple-stream-and-printer-name/make_queue/static/make_queue/js/stream.js) is using. I created this extra attribute in case the name is used for other things.

The Pi's are currently physically labeled as MAKE-001, etc. which should be the websocket url/stream name. But during the transition period, the `stream_name` can be set as the same as the printer name.

### Translations
The `stream_name` form needs to be translated to Norwegian

### Tests
Write javascript tests? Has that been done before in this repo?


make_queue/migrations/0016_auto_20201107_1152.py
```diff
@@ -0,0 +1,29 @@
...
+import web.multilingual.database
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('make_queue', '0015_machinetype'),
+    ]
+
+    operations = [
```
hermabe, line 14: All of these changes seem to be from the MachineTypeField. Is this something that is missing from the previous migration? @ddabble

make_queue/static/make_queue/js/stream.js
```diff
@@ -1,7 +1,7 @@
 function setupSocket($elem) {
     let chatSocket = new WebSocket(
         `wss://${window.location.host}/ws/stream/`
-        + $elem.attr("name").replace(/ /g, "-").replace(/ö/g, "o") + '/');
+        + $elem.attr("data-stream").replace(/ /g, "-").replace(/ö/g, "o") + '/');
```
hermabe, line 4: Since we now have a separate field for the stream name, the `replace()` calls can be removed. It would be better to do some validation in the machine form.

make_queue/templates/make_queue/machine_card.html
```diff
@@ -5,7 +5,7 @@
 <div class="ui {% if color_bottom %}{% card_color_from_machine_status machine %} {% endif %}card">
     <link rel="stylesheet" href="{% static 'make_queue/css/stream.css' %}">
     {% if machine.machine_type.has_stream %}
-        <img class="stream image" name="{% sanitize_stream_name machine %}"
+        <img class="stream image" data-stream="{{ machine.stream_name }}" name="{% sanitize_stream_name machine %}"
```
hermabe, line 8: Is the `name` attribute used anymore? If not can it and the `sanitize_stream_name` tag be removed?

hermabe (changes requested):
Nice! There is no need for the stream to be coupled to the name of the machine.

 - For translations: run `python manage.py makemessages`, update `django.po` and run `python manage.py compilemessages`

make_queue/static/make_queue/js/stream.js
mahoyen: It is there for backwards compatibility, so that the the `name` can be copied over to the `stream_name` in the database, before we manually change the websocket url

hermabe: Wouldn't it be better to set the `stream_name` in the migration so that they automatically are set to the sanitized version?

hermabe:
It would also be nice to have some validation in the machine form so that if you create a machine that has a stream, it must also have a stream name.

make_queue/static/make_queue/js/stream.js
mahoyen: Ah, I am not familiar with migrations, so I did not know that was a possibility. That seems like a better solution. I'll remove the `replace` calls.

hermabe: You can add this at the end of the `operations` list in the migrations:
```python
migrations.RunPython(set_stream_name, migrations.RunPython.noop),
```
and define the conversion function:
```python
def set_stream_name(apps, schema_editor):
    Machine = apps.get_model('make_queue', 'Machine')
    db_alias = schema_editor.connection.alias

    for machine in Machine.objects.using(db_alias).filter(machine_type__has_stream=True):
        machine.stream_name = machine.name.replace("ö", "o").replace(" ", "-")
        machine.save()
```

mahoyen: Where do I define the `set_stream_name` function? When I put it at the end of the "0017_machine_stream_name" file, I get the error:

```
web_1  |   File "/web/make_queue/migrations/0017_machine_stream_name.py", line 18, in Migration
web_1  |     migrations.RunPython(set_stream_name, migrations.RunPython.noop),
web_1  | NameError: name 'set_stream_name' is not defined
```

mahoyen: Moved it above the class, and now it works

make_queue/templates/make_queue/machine_card.html
mahoyen: It is removed

locale/nb/LC_MESSAGES/django.po
```diff
@@ -3,7 +3,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: MAKE NTNU website\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-11-08 22:54+0100\n"
+"POT-Creation-Date: 2020-11-18 21:07+0100\n"
 "PO-Revision-Date: 2018-10-09 14:05+0200\n"
 "Last-Translator: Sindre Stephansen <sindre@sindrestephansen.com>\n"
```
mahoyen, line 8: Should this be updated?

locale/nb/LC_MESSAGES/django.po
```diff
@@ -836,6 +831,14 @@ msgstr "Varighet i timer"
 msgid "Duration in minutes"
 msgstr "Varighet i minutter"
 
+#: make_queue/forms.py:211
+msgid "Stream Name cannot be empty"
+msgstr "Videostrømnavn kan ikke være tomt"
```
mahoyen, line 836: Is "Videostrømnavn" a good translation?

make_queue/forms.py
```diff
@@ -202,6 +202,17 @@ def __init__(self, *args, **kwargs):
...
         )
 
+    def clean(self):
+        cleaned_data = super().clean()
+
+        machine_type = cleaned_data.get("machine_type")
+
+        if machine_type is not None and machine_type.has_stream and cleaned_data["stream_name"] is None:
+            self._errors['stream_name'] = [_("Stream Name cannot be empty")]
+            raise ValidationError(_("stream_name cannot be None"), code="stream_name_is_none")
```
mahoyen, line 212: Not sure if this is best practice, but without line 211, there was no error message when submitting the form. There is also a different message, as a user may not understand the `None` term.

make_queue/forms.py
```diff
@@ -211,7 +222,6 @@ class Meta:
 
 
 class EditMachineForm(BaseMachineForm):
-    machine_type = None
 
     class Meta(BaseMachineForm.Meta):
-        exclude = ["machine_type", "machine_model"]
+        exclude = ["machine_model"]
```
mahoyen, line 227: The machine type is included, so that the javascript that disables `stream_name` works correctly in the `EditMachineView`. Is there a point of excluding `machine_model`? I don't see the harm in leaving it in as well.

make_queue/static/make_queue/js/disable_stream_name_field.js
```diff
@@ -0,0 +1,7 @@
+$(document).ready(function () {
+    $("select[name=machine_type]").change(function () { 
+        var input = document.getElementById("id_stream_name");
+        input.disabled = this.value != 1;
```
mahoyen, line 4: Ideally this should read the `has_stream` property of `machine_type`, but to get it to dynamically change as the choicefield changes, I had to do it in javascript. And I don't think it is possible to get hold of the `has_stream` property in javascript?

make_queue/templates/make_queue/machine/machine_edit.html
```diff
@@ -59,6 +61,13 @@ <h1 class="ui centered header">
                 </div>
             </div>
 
+            <div class="ui two fields">
+                <div class="ui text field">
+                    <label>{{ form.stream_name.label }}</label>
+                    {{ form.stream_name }}
+                </div>
+            </div>
```
mahoyen, line 69: I think that ideally it should not be necessary to specify each field in this way. I believe this can be solved with a for loop and a separate template for form fields. I don't think I should correct it in this PR though.

make_queue/tests/views/test_machine_views.py
```diff
@@ -74,3 +78,32 @@ def create_machine(name_prefix: str, machine_type_: MachineType, priority: Union
...
+        self.response = self.client.get(reverse("edit_machine", args=[machine.pk]))
+        
+        self.assertEqual(self.response.status_code, 200)
+        self.assertTrue(isinstance(self.response.context_data["form"], EditMachineForm))
+
+    def test_createMachine_context_data_has_baseMachineForm(self):
+        self.response = self.client.get(reverse("create_machine"))
+        
+        self.assertEqual(self.response.status_code, 200)
+        self.assertTrue(isinstance(self.response.context_data["form"], BaseMachineForm))
```
mahoyen, line 109: Initially there were some more tests here that checked that `stream_name` is disabled. But as that was solved with javascript, it can't be tested here. The remaining tests are not that useful, so I am wondering if I should just remove them altogether?

mahoyen (commented):
Some points that are up for discussion.

make_queue/migrations/0016_auto_20201107_1152.py
ddabble: Yes, that's correct; I forgot to update the migration after having done quite a few rebases 😅 It was added in #330, however, so this migration should be removed after merging with `dev`.

Maisergodt (changes requested):
Looks good! I have a few comments

make_queue/templates/make_queue/machine/machine_edit.html
Maisergodt: You are welcome to solve it :)

make_queue/forms.py
ddabble: To also add the `code` kwarg to the exception (and also follow Django's preference for using `ValidationError` objects), you can do this: 😊 
```python
raise forms.ValidationError({
    'stream_name': forms.ValidationError(_("Stream name cannot be empty"), code='stream_name_is_none'),
})
```

ddabble:
@mahoyen Are you still active, in the sense that you would be willing to continue to improve this PR..? 🙂

Btw, really sorry that we took so long before we started properly reviewing the PR! (due to being fairly busy at the time it was opened, and kiind of forgetting about it after that 😬)

mahoyen:
I can finish this PR, yes :)

make_queue/models/models.py
```diff
@@ -99,6 +100,18 @@ class Machine(models.Model):
...
+        verbose_name=_("Stream Name"),
+        default=None, blank=True, null=True,
+        validators=[
+            RegexValidator(
+                regex="^[a-zA-Z0-9_-]+$",
+                message=_("Stream name can only consist of english letters, numbers, hyphens or underscores."),
+                code="NOT_URL_SAFE"
+            )
+        ]
+    )
```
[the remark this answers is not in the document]
mahoyen: I would like a custom text, which is why I created my own.

make_queue/static/make_queue/js/disable_stream_name_field.js
mahoyen: Fixed

make_queue/forms.py
mahoyen: Fixed with `self.add_error`

make_queue/static/make_queue/js/stream.js
```diff
@@ -1,7 +1,7 @@
 function setupSocket($elem) {
     const chatSocket = new WebSocket(
         `wss://${window.location.host}/ws/stream/`
```
mahoyen, line 3: When reviewing this, a tip is to change the url manually, so that you can see the streams in action.

```suggestion
        `wss://makentnu.no/ws/stream/`
```

mahoyen:
A fun side effect of this change, is that you can set multiple different machines to show the same stream, if they all contain the same stream name.

make_queue/models/models.py
```diff
@@ -115,6 +116,21 @@ class Status(models.TextChoices):
...
```
Maisergodt, line 131: I don't remember precisely the convention but I think @ddabble introduced a code standard which distinguished between single and double quotation marks. Rougly, you use single in python files and double in the templates. @ddabble is it explicitly defined somewhere? Otherwise there has been som cleanup PRs that address this, like #350 I think.

Maisergodt:
I don't remember precisely the convention but I think @ddabble introduced a code standard which distinguished between single and double quotation marks. Rougly, you use single in .py files and double in the templates. @ddabble is it explicitly defined somewhere? Otherwise there has been som cleanup PRs that address this, like #350 I think.

ddabble:
@Maisergodt Yes, I added it in #376 (see [this section](https://github.com/MAKENTNU/web/blob/093be13343a644f2445e0eb8d69d135bdb7d0f02/CONTRIBUTING.md#string-quotation-marks)) 😊

make_queue/models/models.py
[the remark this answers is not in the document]
ddabble: I would encourage to use `slug_re` (which `validate_slug` uses) as the `regex` kwarg, so that we don't have to redefine a functionally equal regex string 🙂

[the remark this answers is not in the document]
mahoyen: updated!

make_queue/models/models.py
```diff
@@ -115,6 +116,26 @@ class Status(models.TextChoices):
...
+        max_length=30,
+        verbose_name=_("Stream Name"),
+        default=None, blank=True, null=True,
+        validators=[
+            RegexValidator(
+                regex=slug_re,
+                message=_("Stream name can only consist of english letters, numbers, hyphens or underscores."),
+                code='not_url_safe',
+            )
+        ]
```
ddabble, line 135: Some comments on these field options:
* The max length does in principle not need to be constrained, but since it would (most likely) have to be manually copied, we could increase `max_length` to e.g. 50, which is not unwieldy long when manually copying, but still allows for some longer names 🤔
* `null` should generally not be set to `True` on a string-based field (see https://docs.djangoproject.com/en/3.2/ref/models/fields/#null)
* Related to the point above, `default` should not be set to `None`, but the empty string instead
* `verbose_name` should start with a lowercase letter (see https://docs.djangoproject.com/en/3.2/topics/db/models/#verbose-field-names) 😛
* The `RegexValidator` should be moved to `validators.py`, and the `slug_re` regex should be replaced with a regex that disallows capital letters, as allowing both upper- and lowercase letters in a stream name has the potential for unnecessary confusion (as we've previously discussed 😅)

make_queue/forms.py
ddabble: `machine_model` was removed from the `exclude` field in #351 (more specifically d133cd8f9e5b3d41c365b68566b0dd7c59d66c4e), so this is resolved 😄

web/widgets.py
```diff
@@ -21,11 +24,48 @@ class Media:
...
+    """
+    An extension of the ``ChoiceWidget`` (which ``Select`` and ``SelectMultiple`` extends)
+    which allows adding data attributes to each option of the widget's generated ``<select>`` tag.
+    """
+
+    def __init__(self, attr_name_to_attr_value_getter: Dict[str, Callable[[Any], Any]] = None, *args, **kwargs):
+        """
+        :param attr_name_to_attr_value_getter: A dictionary that maps from the name of the data attribute to add,
+                                               to a function that accepts the option's value and returns the value of the data attribute.
+                                               If the function returns ``None``, the data attribute will not be added.
```
Maisergodt, line 37: Nice feature! Is it testable though? (I mean the the custom widget itself)

make_queue/migrations/0023_machine_stream_name.py
```diff
@@ -23,7 +26,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='machine',
             name='stream_name',
-            field=models.CharField(blank=True, default=None, max_length=30, null=True, validators=[RegexValidator(code='NOT_URL_SAFE', message='Stream name can only consist of english letters, numbers, hyphens or underscores.', regex='^[a-zA-Z0-9_-]+$')], verbose_name='Stream Name'),
+            field=models.CharField(blank=True, default='', max_length=50, validators=[django.core.validators.RegexValidator(code='invalid_stream_name', message='This can only consist of lowercase English letters, numbers, hyphens or underscores.', regex=re.compile('^[a-z0-9_-]+$'))], verbose_name='stream name'),
```
Maisergodt, line 29: Does the `message` field need to wrapped inside a translation `_()`?

Maisergodt (commented):
Should there be tests for the customised widgets?

make_queue/models/models.py
ddabble: It's fine that it's not backwards compatible; we'll just change the RPis to use lowercase at the same time as we deploy 🤠🤠

make_queue/migrations/0023_machine_stream_name.py
Maisergodt: Ops, I just realized this was a migration file. I thought this was on the model 😅

Maisergodt (approved):

PR #329, opened 2020-11-10
Use timezone aware datetimes in announcements

hermabe:
When running the tests we get a bunch of RuntimeWarnings: 
```
RuntimeWarning: DateTimeField Announcement.display_from received a naive datetime while time zone support is active.
```

This PR fixes this by using timezone aware datetimes in he `Announcement` model and in `AnnouncementManager`.


ddabble (approved):
Nice! 😄

PR #330, opened 2020-11-13
Feature/secrets

kkmh1807:
Internal secret page:

Created a new content-boxmodel for the internal secrets.

Created 3 new html-files for displaying, editing and creating new secrets.

Created a new css and js file to hide/show secrets.


internal/static/internal/js/secrets.js
```diff
@@ -0,0 +1,13 @@
+const SECRET_SHOW_DURATION_SECONDS = 10;
+
+$(".secret-button").click(function () {
+    const secretButton = $(this);
+    const secretId = secretButton.data("secret-id");
+    const secret = $(`#${secretId}`);
+    secret.removeClass("hidden");
+    secretButton.addClass("hidden");
+    setTimeout(() => {æ
```
sigtheidiot, line 9: I don't think the"æ" is supposed to be there, and it's stopping the timeout from working

internal/templates/internal/secrets_create.html
```diff
@@ -0,0 +1,5 @@
+{% extends 'internal/secrets_edit.html' %}
+{% load i18n %}
+
+{% block title %} {% trans "New secret" %} {% endblock title %}
+{% block form_header %} {% trans "New secret" %} {% endblock form_header %}
```
sigtheidiot, line 5: Here there should be another new line

internal/templates/internal/secrets_edit.html
```diff
@@ -0,0 +1,37 @@
...
+            </div>
+        {% endfor %}
+
+
+        <input class="ui blue right floated submit button" type="submit" value="{% trans "Save" %}"/>
+        <a href="{% url "secrets" %}" class="ui red right floated button">
+            {% trans "Cancel" context "abort action" %}
+        </a>
+    </form>
+{% endblock body %}
```
sigtheidiot, line 37: Another newline missing here

internal/templates/internal/secrets.html
```diff
@@ -0,0 +1,43 @@
...
+                        <i class="red trash icon"></i>
+                    </a>
+                {% endif %}
+                <button class="secret-button ui red button" data-secret-id="secret-{{ secret.pk }}">
+                    {% trans "Press if you're alone" %}
+                </button>
+            </h3>
+            <div id="secret-{{ secret.pk }}" class="hidden">
+                {{ secret.content|safe }}
+                <p>{% trans "Last edited:" %} {{ secret.last_modified|safe }}</p>
```
sigtheidiot, line 39: Maybe change the format of the timestamp, if possible. Right now it will look like  e.g. "2020-11-19 13:28:55.865926+00:00"

sigtheidiot (changes requested):
Very well done, I have left some small comments that you can look at

internal/templates/internal/secrets.html
sigtheidiot: It also shows the time in GMT, and not GMT+1 (which is our timezone)

ddabble: I agree 👍 You don't need the `safe` filter for inserting a datetime, and you can use the `date` filter for formatting the datetime string 🙂 For example:
```suggestion
                <p>{% trans "Last edited:" %} {{ secret.last_modified|date:"d.m.Y H:i" }}</p>
```
(you can look up the format characters in the [documentation on template filters](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#date))

sigtheidiot (approved):

PR #331, opened 2020-11-13
Run apt update before installing system dependencies

hermabe:
Currently the workflow fails when installing system packages because the package index has been changed. This fixes the workflow by running `apt update` before installing system packages.


sigtheidiot (approved):
Thank you 😅

ddabble (approved):
Oh nice, so that's what was causing all these build errors 😅

PR #339, opened 2021-03-03
Reservation error fix

zootos:
A fix for an error that occurs when a reservation is not covered by any rules. Currently, this causes an exception to be raised and sent in an email to the dev team. This fix, simply disallows reservations that are not covered by any rule, as is already the case.

I also included some tests to check these cases specifically, and some missing documentation for the involved methods.


hermabe (approved):
Looks good to me

ddabble (approved):
Nice! 😄 Would be great if you could add a test for when the duration of a reservation is 0, but this should be merged regardless 😊

zootos:
I added the given test. Additionally, I fixed a bug in the template for editing reservations, that caused the datepicker for the start and end time of the reservation to be unusable. Finally, I added some custom error messages when creation of a reservation fails due to either a duration of 0 or not being covered by any rules.

sigtheidiot (approved):

PR #340, opened 2021-03-04
Fixes for course registration search functionality

zootos:
Fixes two issues:

- Properly hides users that do not match search on mobile
- Hitting enter on the keyboard or equivalently the "search" button on a mobile keyboard no longer triggers a download of a spreadsheet of the selected users. Instead, it takes focus away from the search field, i.e., as if the user finished their search. This should be a natural reaction from hitting enter/"search" when searching. For the spreadsheet, the download button must be used, as would be expected.

### DIFF
--- a/make_queue/static/make_queue/css/general.css
+++ b/make_queue/static/make_queue/css/general.css
@@ -40,6 +40,13 @@
     display: none !important;
 }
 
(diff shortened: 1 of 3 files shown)


ddabble (approved):
Great! 🔥

PR #341, opened 2021-03-05
Small features for internal members list

zootos:
Provides some features/fixes for the members list:

- Adds the option to search for users by name (for the given filters)
- Fixes the coloring of the rows (was not working due to the filtering only hiding the non-matching elements, not removing them)
- Only shows the first and last name of the user in the table, rather than all names (the full name is showed in the popup modal after clicking on the user in the table and the full name is used when searching). This should generally allow for more space for the other columns in the table, decreasing the likelihood that the attributes such as committee or status have to be stacked on top of each other.


sigtheidiot (approved):
Looks good!

zootos:
Merge conflicts should be fixed now

PR #342, opened 2021-03-06
Max image size

zootos:
Adds support for specifying a maximum image size. When the image upload widget is used, the input is cleared and the user is shown a simple error message if the image is too large. Currently, the maximum image size is set to 25 MiB, but can easily be adjusted. Either directly in the settings or, better, in the local settings.


sigtheidiot (approved):
Looks good!

zootos:
Merge conflicts should be fixed now.

PR #348, opened 2021-04-12
Fixed missing user info when creating new account 🚑

ddabble:


ddabble:
Merging without approval, as this is pretty critical, and also it's a very simple change that I've already tested works on the [dev domain](https://makentnu.dev/) (by manually changing the files on the server) 😬

PR #350, opened 2021-04-23
More cleanup and fixes

ddabble:
Highlights:
* Refactored member retire/quit to use same view (7e101f1fc6b6b52d5ef8eb73279e055a1c87b758)
* Extended validation of card numbers with checking whether the number is actually the phone number of Building security at NTNU, which seems to happen occasionally, by accident (44c00c748e50fd3ef6eb6885769470f4dfe0853c)
* Cleaned up time formatting, which makes the format of times, dates and datetimes more consistent (5d10039295d9a467eb92a53893c327fcdb5631cc)
* Split [`make_queue/general.css`](https://github.com/MAKENTNU/web/commit/8872a1dc88866503de3ff789d696343a52909418#diff-4ccc3ccb4e8b20602f1a825a345b4d2ff8b90a6ce17b4d349056d628bc561f24) into multiple files (8872a1dc88866503de3ff789d696343a52909418)
* Updated generated gitignore (b2eaf0e2e693b97848d95afad74d125cb65bd1df)
* Improved some code related to Dataporten and LDAP (e904e0c3807de84af733259c1cdfbbc8afa49542, 9e1efe83e1375028c6f71c9f7ad5fa09f638c66e and 84c045339e7dd3bfeba57983a5a51293eb528f86)


internal/forms.py
```diff
@@ -36,24 +36,25 @@ class Meta:
             'committees': SemanticMultipleSelectInput(prompt_text=_("Choose committees")),
         }
 
-    def __init__(self, user, **kwargs):
-        super().__init__(**kwargs)
+    def clean_card_number(self):
```
sigtheidiot, line 39: This is probably corrent, I'm just confused since i cant find a function call for this function in the codebase

sigtheidiot (approved):
This looks good to me! I'm approving, even though I had one comment that i would like to be answered before merging.

internal/forms.py
ddabble: Good question - this is part of some Django "magic" for cleaning specific form fields 😅 See this excerpt from [the docs](https://docs.djangoproject.com/en/3.2/ref/forms/validation/):
> The `clean_<fieldname>()` method is called on a form subclass – where `<fieldname>` is replaced with the name of the form field attribute. This method does any cleaning that is specific to that particular attribute, unrelated to the type of field that it is. [...]

sigtheidiot: Ahh, amazing! Thank you!

PR #375, opened 2021-09-16
Doc/changelog and templates

mahoyen:
## Proposed changes
This PR creates a changelog(with messages based on #statusmeldinger on slack) and pull request templates. 
An issue notice is also added to guide people to our #dev-public channel on slack.
The changelog is in Norwegian, but it should be in English going forward.

"Unreleased" is the title for changes that are not released yet, but will be in the next deployment.

## Areas to review closely
Verify that there are not any information that should not be public in the changelog.

## Checklist
~~- [ ] Created tests which fail without the new change (if possible)~~ Only doc changes.
- [x] All tests passing
- [x] Extended the README / documentation, if necessary
- [x] Extended changelog


ddabble (approved):
Thank you so much for taking the initiative to add these things, they will absolutely be useful! 😊 Also, sorry for taking so long before merging 😅

PR #379, opened 2021-10-15
Fix front page events

ddabble:
The events on the front page are now listed in a "collapsed" fashion, so that events with multiple future `TimePlace`s are listed only once, but with an added label saying how many future occurrences there are (see 48661186e146ef9d16aa45b12a6aef6f96450b70).

Also added a button that appears after the listed events when there are more than 4 future events, which links to the event list page (see f2a0b5e16cbf4e9ccb607058b870e3237c0eaa49 and 8edf416c2d6a29149fea5d531e33340bc3fe50af).

Lastly, added a border between the event list and article list, that only appears on mobile (see 5842224943584232cabf2fe296df8b932a2d32ad).

### DIFF
--- a/web/tests/test_views.py
+++ b/web/tests/test_views.py
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta
+from http import HTTPStatus
 from typing import List
 
 from django.contrib.auth.models import Permission
@@ -147,3 +148,25 @@ class IndexViewTests(TestCase):
         self.assertEqual(len(response_event_dicts), 2)
         self.assertEqual(response_event_dicts[0]['event'], public_event)
         self.assertEqual(response_event_dicts[1]['event'], private_event)
+
+
+class AdminPanelViewTests(TestCase):
+    def setUp(self):
+        self.path = reverse('adminpanel')
+
+    def test_only_users_with_required_permissions_can_view_page(self):
+        def assert_can_view_page(can_view: bool):
+            status_code = self.client.get(self.path).status_code
+            if can_view:
+                self.assertEqual(status_code, HTTPStatus.OK)
+            else:
+                self.assertGreaterEqual(status_code, 300)
+
+        assert_can_view_page(False)
+
+        user = User.objects.create_user(username="user")
+        self.client.force_login(user)
+        assert_can_view_page(False)
+
+        user.user_permissions.add(Permission.objects.get(codename='add_article'))
+        assert_can_view_page(True)


sigtheidiot (dismissed):
I’ve only read over the code and not tested anything on my computer. But from what i’m reading this looks good and can be merged. I can double check that things work, but not until tomorrow.

sigtheidiot (commented):
Okay, i've tested the code in browser now. And a few comments. 

Im not sure of the position and the design of the "More events" link. Looks a bit weird when its on the left side of the events, as i would expect it to be a button on the bottom instead. 

If you mark an event as "frittstående" and then add more occruences, past occurences will be counted on the front page. (This is me using frittstående events wrong btw, but i wanted to test)

(Not important) but i now see that the year does not show up for the events. This is fine for "gjetagende" events and we never really plan events years ahead (or if we do we hide them). But if someone uses the "frittstående" events and makes a new occurence, the old ones will show up and that can be confucing when theres 3 events listed and there is no way to know that the first event was last year. (I know that this is me using "fritstående" evnets wrong btw, therefore not important)

So if this is things we should address, then i would request changes, if not then i'm approving

web/templates/web/index.html
```diff
@@ -33,36 +33,50 @@
...
             {% endfor %}
         </div>
+        {% if more_events_exist %}
+            <h2 class="more-events-link">
+                <a href="{% url 'events' %}">
+                    {% trans "More events" %} <i class="angle double right icon"></i>
+                </a>
+            </h2>
+        {% endif %}
+        <hr class="make-bg-yellow event-article-border"/>
```
Maisergodt, line 79: Isn't the yellow make color already specified in `event-article-border` which makes it redundant?

ddabble: Well, the `event-article-border` class sets `border-color`, whereas `make-bg-yellow` sets `background-color`, so I wouldn't say it's redundant 😅

ddabble:
@sigtheidiot Nice changes! I worked a bit further on your improved design (and some other improvements), so do let me know what you think! 😊 

---

> If you mark an event as "frittstående" and then add more occruences, past occurences will be counted on the front page. (This is me using frittstående events wrong btw, but i wanted to test)

Yes, that part of the code was based on the code in [`ViewEventsView`](https://github.com/MAKENTNU/web/blob/48661186e146ef9d16aa45b12a6aef6f96450b70/news/views.py#L36-L47), and it seems like this behavior is intentional 🙂 

> But if someone uses the "frittstående" events and makes a new occurence, the old ones will show up and that can be confucing when theres 3 events listed and there is no way to know that the first event was last year.

True, but that would be a case of the user misusing the feature, which I don't think should warrant changing a thing that assumes a correct usage of the feature 😛

sigtheidiot (approved):
This looks good to me now!

web/templates/web/index.html
Maisergodt: Ah, then it seems that the `background-color` minimally impact the line though.

Maisergodt (approved):

PR #396, opened 2021-11-25
Make dirtied forms prevent user from leaving

ddabble:
This adds a library [jquery.dirty](https://github.com/simon-reynolds/jquery.dirty) and a script (in [common_utils.js](https://github.com/MAKENTNU/web/blob/e8a06e438496b4f9030dfb62ce4e507ea712f089/web/static/web/js/common_utils.js#L9-L13)) which make all forms on the site prompt users before leaving a page, when a form on that page has been "dirtied" (modified).
*Individual forms can opt out of this by adding the `dont-prevent-leaving` class.*

### DIFF
--- a/web/static/lib/jquery/jquery.dirty.js
+++ b/web/static/lib/jquery/jquery.dirty.js
@@ -1,0 +1,308 @@
+/*
+ * Dirty
+ * jquery plugin to detect when a form is modified
+ * (c) 2016 Simon Taite - https://github.com/simon-reynolds/jquery.dirty
+ * originally based on jquery.dirrty by Ruben Torres - https://github.com/rubentd/dirrty
+ * Released under the MIT license
+ */
+
+(function($) {
+
+    //Save dirty instances
+    var singleDs = [];
+    var dirty = "dirty";
+    var clean = "clean";
+    var dataInitialValue = "dirtyInitialValue";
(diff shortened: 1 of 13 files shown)


web/static/lib/jquery/jquery.dirty.js
```diff
@@ -0,0 +1,308 @@
...
+
+(function($) {
+
+    //Save dirty instances
+    var singleDs = [];
+    var dirty = "dirty";
+    var clean = "clean";
+    var dataInitialValue = "dirtyInitialValue";
+    var dataIsDirty = "isDirty";
+
```
Maisergodt, line 17: Is there any reason why you are using `var` instead of  `let` or `const`? If not, then I think it is preferable to avoid using `var` 

(see https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var)

Maisergodt (changes requested):

web/static/lib/jquery/jquery.dirty.js
ddabble: I absolutely agree, but this is a library I downloaded (see the comment at the top of the file), and I generally don't think we should put resources into cleaning up those 😅

Maisergodt (approved):

PR #400, opened 2022-01-07
Fix article image heights + failing tests

ddabble:
* 🚑 Fixed article images having 0 height (d8b0ce8601d061f92ecf9c7d9d4ed2ce54270889)
* Improved some code related to the `robots.txt` files (f43c67968fd4447394da17eba95d785927e2d3bb, 590e1830f6ed57895791b98d01cf77bf6534e56c, 1e01ca4d05964f311c26d0d2119b5da154413811)
* Fixed failing tests by replacing `>=` with `==` in `requirements.txt` (fb9332c7a47152055dbcc4253357e2784dde1f69)
* Updated some requirements (25ade25cb6895ed78dc44e7acdf0a1943405d0ec)


internal/templates/internal/robots.txt
```diff
@@ -1,2 +1,2 @@
 User-agent: *
-Disallow: /*?
+Disallow: /
```
Maisergodt, line 2: What is the difference of letting the URLs being crawled here vs e.g. `web/`?

Maisergodt (approved):
Just a general question regarding disallowing web crawlers to crawl on certain URLs, other than that it looks good to me :)

internal/templates/internal/robots.txt
ddabble: Hm, do you mean if it said `Disallow: web/` instead of what it currently says..? 🤔

ddabble:
These tests should also succeed after #400 has been merged 🤠

internal/templates/internal/robots.txt
Maisergodt: Oh, I meant you added a disallowed: /*? at web/robots.txt and was curious by you added/removed the asterisk on certain pages. Hope it makes the question clearer :)

ddabble: Ah, I see 😅 So:
* in `internal/robots.txt`, I set `Disallow` to just `/`, as we should disallow crawling of *all* the pages of that subdomain (as it's not public), instead of only disallowing crawling of URLs with query parameters;
* in `admin/robots.txt`, the reason is the same as for `internal`;
* in `web/robots.txt`, I added `Disallow: /*?`, because we generally don't want crawling of query URLs, as they're most often created on a whim (as part of searching/filtering/sorting things on a page), and not something we want appearing among the search results of a search engine;
* in `docs/robots.txt`, I left the `Disallow` rule as it is, for the same reason as for `web`.

Did this clear things up? 😊

PR #409, opened 2022-01-12
Make fonts self-hosted

ddabble:
This makes us self-host the Ubuntu font that we're using (b72011d8789a6e1da9428367be7a658c8ebc13cf), and adds a math formula button to CKEditor with also adds some additional self-hosted fonts for rendering the math formulas (bd613e230e5fd45bea91eee2181b7f1ff94ddbea).


maritmoe (commented):
Self-host the Ubuntu font is working. 
The math formula does not appear as a formula in the finished article, the TeX code is displayed instead.

By adding:
`<script type="text/javascript" async
        src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_CHTML">
</script>`
to the news_base_detail.html the formulas appear as formulas.
I tried to make a commit with the change I did. I do not know if it is okay to do it like i did.

ddabble:
@maritmoe Excellent, thank you for figuring out a solution, and for that code snippet! 😄 I implemented it in 058cf850aaf92910b0da85dba8a2d339d2176010 (but with locally hosted scripts 😋).

> I tried to make a commit with the change I did. I do not know if it is okay to do it like i did.

Do you mean that you made a commit locally, but did not push it yet..? If so, did it contain any additional improvements compared to the implementation in 058cf850aaf92910b0da85dba8a2d339d2176010? 🙂

In any case, if the functionality and the code now looks good, would you mind leaving an approval review? 😊

maritmoe (approved):
The math formula functionality now works as intended :)

PR #411, opened 2022-01-12
Fix some security issues

ddabble:
This fixes the following issues:
* [Incomplete multi-character sanitization](https://github.com/MAKENTNU/web/security/code-scanning/48) in `mazemap_search.js` (2cfa80c72ffde644dcf9fa73552a60533d546393); this is unlikely to be exploited, as the unsanitized data comes directly from the MazeMap API
* [Reflected server-side cross-site scripting](https://github.com/MAKENTNU/web/security/code-scanning/20) in `RFIDView` and subclasses (f826108f52a49553670f3b9c4104fcf3dabc8c71); this is highly unlikely to be exploited, as it's only possible to do so for the very few people who have access to the `CHECKIN_KEY` setting

### DIFF
--- a/card/views.py
+++ b/card/views.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.http import HttpResponse
 from django.utils.decorators import method_decorator
+from django.utils.html import escape
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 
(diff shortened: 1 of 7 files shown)


ddabble:
Merging without waiting for approval, so that the security vulnerabilities are patched.

PR #428, opened 2022-03-19
Some event ticket improvements

ddabble:
## Main changes:
* **Removed `EventTicket`'s legacy `_name` and `_email` fields (17cb55dfdfaca11fb7b5c70ac0ebe6df653f90ad)**
  * It's been a few years since we started requiring being logged in when registering for an event, and so it's definitely time to remove these backup fields. (If it would ever be useful to dig up the ticket holders of old events, we have backups of the database from back then anyway 🤠)
* **Event tickets are reactivated instead of recreated (9849d1467d32173af6ffd7810bba6783684672d8)**
  * Previously, when non-admin users would re-register for an event after having canceled their tickets, a new, active ticket would be created, but the old, canceled ticket would remain in the database. Keeping the canceled tickets like this, is currently not being utilized by anyone in the organization, and so we stop needlessly cluttering the database with these objects.

## Additional changes:
* Made users with the `news.cancel_ticket` permission always able to register for an event (6bc15bb7e2dcbd0d6cc98fa807da80eb4f745e87)
* Improved event ticket card templates, like displaying the ticket's language and comment (73c5e1a2285c2b3216fb8466e14b8e506c1c3b15)


maritmoe (approved):
Had some trouble running migrate after checking out the branch locally. But it worked after I deleted some tickets from the database that had both event and timeplace connected to it. Can not find any problems with the code. The comment and language from the first registration is shown when you try to register again after you have cancelled a ticket. Old canceled ticets are being deleted when a user reactivates the ticket. Nice work!

PR #620, opened 2023-03-27
Add custom `makemessages` command

ddabble:
## Proposed changes

Override the `makemessages` management command in order to do the following:
- Inject default `--ignore` options (0aea8e0b084feb9887486fe6f7142b91504316c1)
  - This is to avoid having the command iterate over files outside the Django source directory (`src`), like inside the `node` directory - which might contain a `node_modules` directory containing tens of thousands of files
- Make it automatically convert the path format of the file location comments in `.po` files generated on Windows, to the format generated on Linux - i.e. forward slashes (`/`) instead of backslashes (`\`) and no leading `.\` (1b4685eb2a9b668724b706727979f900799d461a)
  - In other words, developers using Windows no longer have to do this manually after running the command :)

## Areas to review closely

That the `makemessages` command produces the proper output.

## Checklist

- [ ] Created tests that fail without the changes, if relevant/possible
- [x] Made sure that your code conforms to [the code style guides](https://github.com/MAKENTNU/web/blob/main/CONTRIBUTING.md#code-style-guides) and that any [common code smells](https://github.com/MAKENTNU/web/blob/main/CONTRIBUTING.md#code-review-guideline-code-smells) have been addressed
  - (It's not intended that you read through this whole document, but that you get yourself an overview over its contents, and that you keep it in mind while taking a second look at your code before opening a pull request)
- [x] Added sufficient documentation - e.g. as docstrings or in the [README](https://github.com/MAKENTNU/web/blob/main/README.md), if suitable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ A summary of changes made to the codebase, grouped per deployment.
 - Fixed not loading a machine's page when using the keyboard to select a different machine in the dropdown on a machine detail page
 - Fixed JavaScript errors when changing the machine or machine type in the reservation creation form
 - Added GitHub Actions workflow for automatically labeling deployment PRs
+- Made the `makemessages` management command automatically convert the path format of the file location comments in `.po` files generated on Windows,
+  to the format generated on Linux - i.e. forward slashes (`/`) instead of backslashes (`\`) and no leading `.\` - so that developers using Windows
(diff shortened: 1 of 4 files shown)


ddabble:
Merging without explicit approval from another member, as the Dev committee agreed to merge these changes and assume they're relatively bug-free, simply to get things done quicker.

PR #669, opened 2023-08-17
Added page to show statistics from use of the workshop

elisakiv:
## Proposed changes

Created a statistics page to view statistics regariding the use of the workshop, such as reservations made, and when the workshop is in use.

## Checklist


Gunvor4:
I find "Totalt hvor lenge hver symaskin har blitt reservert" a bit complicated. I would suggest writing this in a simpler way, for example "Total reservasjonstid for 3D-printere" in Norwegian and "Total reservation time for 3D printers" in English. Same goes for sewing machines.

Gunvor4:
I also find "Antall ganger hver 3D-printer har blitt reservert" a bit complicated. Maybe we could write "Antall reservasjoner pr. 3D-printer" (Norwegian) / "Number of reservations pr. 3D-printer" (English). Same goes for sewing machines.

Gunvor4:
"Top 3 lengste 3D-printer reservasjoner" has a grammatical error. It should be "3D-printer-reservasjoner", with a hyphen. But this could also be rewritten to make it simpler. Suggestion "Topp 3 mest reserverte (evt. populære) 3D-printere" (Norwegian) / "Top 3 most reserved (or popular) 3D-printers" (English). Same goes for sewing machines.

Gunvor4:
On the third plot, some of the sewing machines appear twice
Bug_symaskindiagram_2

Gunvor4:
On a normal screen, with two plots next to each other, the title is above the second plot. It would look better if it was in the middle. The two plots are also not situated in the middle of the page, which is fine, but then the title of the page should be adjusted relative to the plots, and not the page. The same issue is still there when the screen size gets smaller and only one plot is shown per row.

Gunvor4:
In the two plots counting how many times a 3D printer or a sewing machine has been booked by someone, it looks a bit strange that there are decimals on the y-axis, since it is impossible for a machine to be booked 0.5 times. The two other plot, on the other hand, are counting hours, so there it would have been ok to see decimals. But these plots only have integers on the y-axis. The styles should be switched – the "number of times" plots should not have decimals, and the "how many hours" plots, could.

Gunvor4:
The last plot doesn't show any numbers on the y-axis, and the axis also doesn't have a label. I guess what is shown here is the result of a calculation? Still – would it be possible to give the y-axis a label or introduce some numbers here, to make it clearer what the plot actually shows? Maybe add a label like "equipment currently in use" or "machines or workstations currently in use"?

Gunvor4:
On the x-axis on the last plot – the numbers are tilted, like for the names of 3D-printers or sewing machines. But on this plot, the only thing the x-axis shows are time points, which are shorts, so tilting them looks a bit strange.
Tilt_x-axis

src/makerspace/templates/makerspace/statistics.html
```diff
@@ -0,0 +1,68 @@
...
+    <script defer src="{% static 'makerspace/js/statistics_charts.js' %}"></script>
+{% endblock extra_head %}
+
+{% block body%}
+<div class="ui container">
+    <h1 class="ui text-centered make-col-blue header"> {% translate "Statistics" %}</h1>
+    <div class="ui column cards">
+    <div class="card " style="height: 50vh; width: 65vh">
+            <div class="content">
+            <div class="header">{% translate "How long each 3D-printer* has been reserved in total:" %} </div>
```
Gunvor4, line 18: Maybe the plots would look cleaner if instead of showing the asterisk text at the bottom of the plots, the word "3D-printer*" was instead put in a <span>-element, and the asterisk text included in the title attribute. Then, the text would only be shown if someone is hovering over the word "3D-printer*". To make it visible that this word was different, the span element could be styled differently than the rest of the text (not necessarily the same way as I did it in the example, because that looks ugly), and also given a pointer cursor. But this is only a suggestion – not something that absolutely needs to be changed.
```suggestion
            o	<div class="header"> {% blocktranslate %}Number of times each <span title="Inkluderer 3D-printere, Raise3D-printere og SLA 3D-printere" style="background: yellow; cursor: pointer;">3D-printer*</span> has been reserved:{% endblocktranslate %} </div>
```

Gunvor4, line 18: I find "Totalt hvor lenge hver symaskin har blitt reservert" a bit complicated. I would suggest writing this in a simpler way, for example "Total reservasjonstid for 3D-printere" in Norwegian and "Total reservation time for 3D printers" in English. Same goes for sewing machines.

src/makerspace/templates/makerspace/statistics.html
```diff
@@ -0,0 +1,68 @@
...
+            {{ span_of_printer_reservations|json_script:"span_of_printer_reservations" }}
+            <canvas id="span_printer_reservations" width="800" height="550">
+            </canvas>
+            </div>
+        </div>
+
+    <div class="card" style="height: 50vh; width: 65vh">
+            <div class="content">
+            <div class="header"> {% translate "Number of times each 3D-printer* has been reserved:" %} </div>
```
Gunvor4, line 27: I also find "Antall ganger hver 3D-printer har blitt reservert" a bit complicated. Maybe we could write "Antall reservasjoner pr. 3D-printer" (Norwegian) / "Number of reservations pr. 3D-printer" (English). Same goes for sewing machines.

src/makerspace/templates/makerspace/statistics.html
```diff
@@ -0,0 +1,68 @@
...
+            <div class="content">
+                <div class="header"> {% translate "Number of times each sewing machine has been reserved:"%} </div>
+                {{ number_of_sewingmachine_reservations|json_script:"number_of_sewingmachine_reservations" }}
+            <canvas id= "sewingmachine_reservations" width="800" height="550"></canvas>
+            </div>
+        </div>
+
+        <div class="card " style="height: 50vh; width: 65vh">
+            <div class="content">
+                <div class="header">{% translate "Top 3 longest 3D-printer* reservations:"%}  </div>
```
Gunvor4, line 51: "Top 3 lengste 3D-printer reservasjoner" has a grammatical error. It should be "3D-printer-reservasjoner", with a hyphen. But this could also be rewritten to make it simpler. Suggestion "Topp 3 mest reserverte (evt. populære) 3D-printere" (Norwegian) / "Top 3 most reserved (or popular) 3D-printers" (English). Same goes for sewing machines.

src/makerspace/static/makerspace/js/statistics_charts.js
```diff
@@ -0,0 +1,297 @@
...
+
+    new Chart(ct1, {
+            type: 'bar',
+            data: {
+                labels: data.map((element)=> element.name),
+                datasets: [{
+                    label: "How long each printer has been reserved in total",
+                    data: data.map((element)=> (element.len)),
+                    backgroundColor: [
+                        'rgba(249, 65, 68)',
```
Gunvor4, line 18: -	I like the colors of the plots. Still, it might be more logical if they were MAKE's official colors, which are MAKE yellow (#F8C811), dark blue (#0C426A) and turquoise (#00EBB3), in addition to two different shades of gray - #4A4B4C and #DCDDDE. We also have an orange variation of the yellow (#F5A13F). I've created some examples. The colors of the gradient were generated using this tool (https://coolors.co/gradient-palette/eeddf3-ee92b1?number=7)
If you decide to keep the colors you already have, maybe the rainbow plot could be smoother, because in the second and fourth bar looks more orange, while the third and the fifth looks more yellow. Also – the ninth bar looks more purple and the tenth looks more blue. On the other hand - the other plots seem to follow a gradient.
Also – if you have 10 colors in your gradient, and we have more than 10 3D printers, we would go from the 10th color and back to the first, which will not look too nice. Maybe the gradient could be calculated in the JavaScript file, from the first and the last color, and the number of bars? If not, it would also be possible to reverse the gradient, as I've shown in the example.

```suggestion
                        '#00EBB3',
                        '#03C1A1',
                        '#06978F',
                        '#096C7C',
                        '#0C426A',
                        '#096C7C',
                        '#06978F',
                        '#03C1A1',
```

src/makerspace/views.py
```diff
@@ -67,3 +70,48 @@ class EquipmentDeleteView(PermissionRequiredMixin, PreventGetRequestsMixin, Dele
...
     model = Equipment
     success_url = reverse_lazy('admin_equipment_list')
+
+
+class StatisticsView(TemplateView):
+    template_name = 'makerspace/statistics.html'
+    get_time_in_hours = 3600000000  # number the reservation length needs to be divided by to get the length in hours
+    filter_reservations = Q(reservations__special=False) & Q(reservations__event__isnull=True)
+
+    reservations = Reservation.objects.prefetch_related(Prefetch('machine', queryset=Machine.objects.
```
Gunvor4, line 80: The exprassions for reservations (line 80 – 82) sewingmachines (line 84 – 87) and printers (line 89 – 92) are extremely long and ugly. Would it be possible to refactor these, to make them more readable?

src/makerspace/templates/makerspace/statistics.html
```diff
@@ -0,0 +1,68 @@
...
+
+{% block extra_head %}
+    <script defer src="{% static 'makerspace/js/statistics_charts.js' %}"></script>
+{% endblock extra_head %}
+
+{% block body%}
+<div class="ui container">
+    <h1 class="ui text-centered make-col-blue header"> {% translate "Statistics" %}</h1>
+    <div class="ui column cards">
+    <div class="card " style="height: 50vh; width: 65vh">
```
Gunvor4, line 16: Generally on the MAKE website, styles are put in separate CSS documents. Maybe that should be done here as well, instead of styling each element with the same CSS code.

src/makerspace/templates/makerspace/statistics.html
Gunvor4, line 18: Since the elements on line 18 - 21 are inside of a div-element, they should probably be indented. Same goes for lines 27 - 29, 35 - 37 and 45

Gunvor4:
I don't think you need a colon after the title of a plot

Gunvor4:
It might be an idea to sort the printers, either alphabetically, or to keep the same order as they appear in on the website. Now I added some fake ultimakers, then some fake raise3D printers, a fake SLA printer, and then some fake ultimakers again, and now I have the ultimakers at both sides of the plot, with raise 3D printers and SLA printers in the middle.

Gunvor4:
Aaaand now I'm done! :-) Sorry about being nitpicky :-P

ddabble (changes requested):
Amazing job! Properly displaying statistics is usually a pretty complex task, so I'm genuinely impressed by what you've created 😊 However, complexity also necessarily increases the risk of overlooking smaller details, and so I've found quite a few things to provide feedback on 😅 If you'd like some help making any of the suggested changes, or if you'd simply like to offload some work, I'd be happy to help :)

---

First off, it seems like no parts of the code you've added use any of the things in the `makerspace` app, and instead only use things in the `make_queue` app... I guess it makes sense you added it to the former app since all of the statistics relate to the physical makerspace (another reason why the two mentioned apps have pretty terrible names, in my opinion 😅), but the code should really be moved to the `make_queue` app. This will require a little work, but I'd absolutely recommend you do this, for future maintainability 🙂

It would be great if you could fix [the issues detected by Code Climate](https://codeclimate.com/github/MAKENTNU/web/pull/669) (link is in the status checks at the bottom of this PR page) 🙂 (It should be mentioned that fixing auto-detected issues like those, might not necessarily make the code any better - it might even make the code worse in some cases - but they're usually a very good indication!) It looks like most of the issues can/will be fixed after addressing some of my comments below, though :)

Lastly, it would be fantastic if you could add some unit tests for the view code, to both make sure that it works the way we think it does, and to provide concrete examples for future developers, which can help them understand the intended function of the code 😊

Btw, well done minimizing the number of DB queries using prefetching! That's often quite a daunting task, but it looks like you've implemented it well! 😄

src/makerspace/views.py
```diff
@@ -67,3 +70,48 @@ class EquipmentDeleteView(PermissionRequiredMixin, PreventGetRequestsMixin, Dele
...
+        return dict(time.items())
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        span_of_printer_reservations = list(self.get_machines_with_reservations("3D-printere").
+                                            annotate(len=Sum((F('reservations__end_time') - F('reservations__start_time')) / self.get_time_in_hours, output_field=IntegerField())).values('id', 'name', 'len'))
+        number_of_printer_reservations = list(self.get_machines_with_reservations("3D-printere").annotate(number_of_reservations=Count('reservations')).values('id', 'name', 'number_of_reservations'))
+        longest_printer_reservations = list(self.get_machines_with_reservations("3D-printere").
+                                            annotate(len=Sum((F('reservations__end_time') - F('reservations__start_time')) / self.get_time_in_hours, output_field=IntegerField())).
+                                            order_by('-len').values('id', 'name', 'reservations', 'len'))[:3]
```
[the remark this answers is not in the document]
ddabble: Oh wait, after testing some of the code locally, it seems like `longest_printer_reservations` does indeed contain what the name suggests, and the annotated `len` attribute's value is, appropriately, the duration of each reservation - in contrast to [the renaming suggestion I made in a comment above](https://github.com/MAKENTNU/web/pull/669#discussion_r1403761054) 😅 The code makes this pretty confusing, since each row/element of the list does _not_ represent each machine - the opposite of what you'd expect, and what the lists above and below contain. _(I'm explaining this to avoid any potential misunderstandings:)_ Instead, due to the way `values()` works, when providing the name of a many-to-many field or a reverse many-to-one manager (`'reservations'` in this case), it will return one row/element for each combination the multivalued fields provided (one row per reservation, in this case), as explained in the warning at the bottom of [the `values()` section in the docs](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#values). Under [the `values_list()` section](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#values-list), this is mentioned:
> **`values()`** and **`values_list()`** are both intended as optimizations for a specific use case: retrieving a subset of data without the overhead of creating a model instance. This metaphor falls apart when dealing with many-to-many and other multivalued relations (such as the one-to-many relation of a reverse foreign key) because the “one row, one object” assumption doesn’t hold.

So, I think code relying on such an obscure fact of the `values()` method should be avoided (or at the very least explicitly pointed out and explained in comments). I think I found a fairly easy replacement, however, which also makes the query slightly less complex by going through the reservation table instead of the machine table: 🙂(This contains some elements from other parts of the code and from my other suggestions, as you can see :))
```python
longest_printer_reservations = (
    Reservation.objects.filter(event=None, special=False, machine__machine_type__pk=1)
    .annotate(duration=Sum(F('end_time') - F('start_time')))
    .order_by('-duration')
    .select_related('machine')
    [:3]
)
longest_printer_reservations_list = []
for reservation in longest_printer_reservations:
    duration: timedelta = reservation.duration
    longest_printer_reservations_list.append({
        'machine_name': reservation.machine.name,
        'duration_hours': duration.total_seconds() // (60 * 60),
    })
```
`longest_printer_reservations_list` can then be inserted into the context variables 🙂 Any thoughts?

PR #717, opened 2024-02-23
Remove Discord username validation

Gunvor4:
## Proposed changes

Removed the regex validating Discord usernames, since Discord somewhat recently changed their username format (see https://discord.com/blog/usernames and https://support.discord.com/hc/en-us/articles/12620128861463-New-Usernames-Display-Names). Also removed the `help_text` explaining the old username format.

## Checklist


elisakiv:
The help text also need to be updated in the member model, the discord field in the Django admin panel still asks for the username to include the hashtag and the numbers

elisakiv (approved):

src/internal/validators.py
```diff
@@ -34,8 +34,8 @@ def validate_semester_float(value):
         )
 
 
-discord_username_regex = re.compile(r'^(.+)#([0-9]{4})$')
+discord_username_regex = re.compile(r'^(?!.*\.\.)[a-z0-9_.]{2,32}$')
```
ddabble, line 37: To repeat the sentiment of [this comment](https://github.com/MAKENTNU/web/blob/70270b1eb6e69f2d2dfd99e84d179f17a6483e34/src/internal/models.py#L58), I don't think we should write code that keeps track of these minute details of other services as long as it doesn't affect us much 🤔 The original regex was mainly there to help users input the correct value, in case they forgot to include the four digits at the end (which I think used to be a 
fairly common problem), and because it was such a simple regex string (notice in particular the previous `(.+)`, which matched the whole username except the hashtag and last four digits). However, I think the new regex - which I'm guessing is probably pretty much the exact same as Discord themselves use (?) - is unnecessary detailed for our use case; why would we care exactly which characters and how many that the usernames consist of?

So I think it's best to simply remove the Discord username validation altogether, including [the `validators` argument](https://github.com/MAKENTNU/web/blob/70270b1eb6e69f2d2dfd99e84d179f17a6483e34/src/internal/models.py#L60) 🙂

Gunvor4:
> Did you remember to run `makemigrations`? These changes to the model will necessarily create a new migration :p None of the changes affect the database schema in any way, though, so I think it's better to simply remove `help_text` (and `validators`, re: my comment above) from the `internal` migrations 0015 and 0024 - which are the only ones that include those arguments of the `discord_username` field 🙂
> 
> We can also remove the translation strings for the previous `help_text` (and `discord_username_validator`) from `django.po`; recompiling `django.mo` shouldn't be necessary, though it won't hurt :)

Done and done :-)

src/internal/migrations/0027_alter_historicalmember_discord_username_and_more.py
ddabble, line 1: Nice! Though as I mention in [my comment](https://github.com/MAKENTNU/web/pull/717#issuecomment-1967011114):
> None of the changes affect the database schema in any way, though, so I think it's better to simply remove `help_text` (and `validators`, re: my comment above) from the internal migrations **0015** and **0024** [instead of creating an additional migration]

Introducing an extra migration doesn't hurt much at all, but it builds up over time (e.g. the creation of a test database when running tests in debug mode takes 10+ times as long now as it did a few years ago), so better to avoid adding one if not necessary 🙂 (This can of course be "undone" by [squashing the migrations](https://docs.djangoproject.com/en/stable/topics/migrations/#squashing-migrations) :))

Gunvor4: I now made a squashed migration file. I had to transfer some functions from other migration files manually, so please take a look and check if I did it correctly.

ddabble: Woahh, I like your initiative for fixing things, but this is wayy out of the scope of this PR 😅 Highly recommend that you open a new PR that's solely focused on squashing the migrations, as that will be a fairly big job to review 🙂

(On that last note, when squashing the migrations, we should do so for *all* the apps, not just one of them. Also, the migrations should be heavily revised, as the auto-generated ones will contain lots of unnecessary code, like custom-written functions passed to `RunPython`, that are completely unnecessary to keep in the squashed one (think of it like a new `0001_initial` migration).)

ddabble: Btw, if it's unclear what I meant by the part of my original comment that I quoted above, I can just push a commit, if it's okay for you :) It'll be a very small change.

Gunvor4: I read your original comment as though you wanted me to squash. But sure - if that was _not_ what you meant after all - go for it.

ddabble: Ah, I see 😅 What I meant was that, in general, we should only add migrations when necessary; otherwise, we should do small changes to the existing migrations that circumvent generating a migration when running `makemigrations` - like the changes I mentioned above. And in the long run, we should ideally squash the existing migrations somewhat periodically - e.g. every 5-10 years or when running tests in debug mode takes an unbearably long time :)

PR #754, opened 2025-02-27
Fix quote sort

TheStrgamer:
## Proposed changes

Made it so quotes that have the same date are sorted by descending id.

## Checklist


src/internal/views.py
```diff
@@ -286,7 +286,7 @@ class QuoteListView(ListView):
     model = Quote
     template_name = 'internal/quote_list.html'
     context_object_name = 'quotes'
-    queryset = Quote.objects.order_by('-date').select_related('author')
+    queryset = Quote.objects.order_by('-date', '-id').select_related('author')
```
ddabble, line 289: Sorting by ID is very unusual, as the only/main purpose of an ID is for it to be unique, and not e.g. have a particular ordering. It's not impossible for the table's auto-increment counter to reset (though this must usually be done manually), which is hard to detect through code, or [changes to the database design might break this](https://stackoverflow.com/a/3731048).

So I think that a "proper" solution would be to either:
1. Change `date` to be a `DateTimeField` and let the users optionally choose a time in addition to the date (though this might be a bit weird to use)
1. Add a new `ordering` or `priority` field (like `Machine` and `MachineType`) and sort on that
1. Add a new `created` field (with `auto_now_add=True`) or `last_modified` field (with `auto_now=True`) and sort on that

That being said, this solution will absolutely most likely work for the foreseeable future, so if you'd prefer to not spend more time on this at all, it's understandable :)

ddabble (approved):
Looks good! 😊

ddabble (commented):
Ah, I forgot: I think you should update the changelog with this fix :)

